PR #1, opened 2023-08-03
PRMDR18 Create upload document url lambda 

NogaNHS:


SRAlexander (changes requested):
I'm happy with this from a code perspective :)

Could we however move the lambda into it's own folder? 

- lambdas
    - Upload Document Lambda
    
This will enable set folder builds with GitHub actions so we only rebuild functions that have explicit changes. Cheers :)

PR #13, opened 2023-08-22
PRMDR-80: Migrate patient search lambda and project refactor

abbas-khan10:


Makefile
```diff
@@ -0,0 +1,83 @@
...
+	find . -name '*.pyo' -exec rm -f {} +
+	find . -name '__pycache__' -exec rm -fr {} +	
+
+clean-test:
+	find . -name '.pytest_cache' -exec rm -fr {} +
+	find . -name '.coverage' -exec rm -fr {} +
+	find . -name 'htmlcov' -exec rm -fr {} +
+	find . -name '.cache' -exec rm -fr {} +
+
+format:
```
SRAlexander, line 21: These require a venv to run, shold we have some sort of check to make sure it exists before running the rest of this command + others below?

Makefile
```diff
@@ -0,0 +1,83 @@
...
+	./lambdas/venv/bin/python3 -m isort lambdas/services/. 
+	./lambdas/venv/bin/python3 -m isort lambdas/tests/. 
+
+test-unit:
+	cd ./lambdas && ./venv/bin/python3 -m pytest tests/
+
+test-unit-collect:
+	cd ./lambdas && ./venv/bin/python3 -m pytest tests/ --collect-only
+
+env:
```
SRAlexander, line 36: Ignore previous comment, I've spotted this and the comment in the readme. :)

Makefile
```diff
@@ -0,0 +1,83 @@
...
+	rm -rf lambdas/venv || true
+	python3 -m venv ./lambdas/venv
+	./lambdas/venv/bin/pip3 install --upgrade pip
+	./lambdas/venv/bin/pip3 install -r lambdas/requirements.txt
+	./lambdas/venv/bin/pip3 install -r lambdas/requirements-test.txt
+
+zip:
```
SRAlexander, line 43: Looks good but we may need a chat on how we deploy all lambdas at once vs individually.

README.md
```diff
@@ -1 +1,26 @@
...
+https://github.com/pyenv-win/pyenv-win
+
+
+### Virtual Environment
+
+To setup the Python environment for backend development, run: `make env`
+
+This will create a virtual environment with all production and test requirements. The virtual environment can be found at `.lambdas/venv`.
+
+To activate the environment run `source ./lambdas/venv/bin/activate`
```
SRAlexander, line 21: I believe this varies between OS, could we add a command for Mac and PC?

lambdas/services/pds_patient.json
```diff
@@ -0,0 +1,411 @@
+{
```
SRAlexander, line 1: Since this is a mock / expected result for the service at this point in time, should we keep these in a separate folder? e.g. services/mocks/pds_patient.json?

lambdas/services/pds_patient_restricted.json
```diff
@@ -0,0 +1,83 @@
+{
```
SRAlexander, line 1: Same for this data.

SRAlexander (changes requested):
Solid first review. :)
There are a couple of discussion points but for the most part it looks good :)

Makefile
```diff
@@ -0,0 +1,85 @@
...
+	./lambdas/venv/bin/python3 -m isort lambdas/services/. 
+	./lambdas/venv/bin/python3 -m isort lambdas/tests/. 
+
+test-unit:
+	cd ./lambdas && ./venv/bin/python3 -m pytest tests/
+
+test-unit-collect:
+	cd ./lambdas && ./venv/bin/python3 -m pytest tests/ --collect-only
+
+env:
```
joefong-nhs, line 36: could we use the same `make env` and `make test-unit` in our github action? 
I see that in github action "Test with pytest", 
we are not making venv, and we seems not running pip install with `requirements-test.txt`

Makefile
```diff
@@ -0,0 +1,85 @@
...
+.PHONY: Clean install
+clean-install:
+	cd ./app && npm ci
+
+.PHONY: Start
+start:
+	cd ./app && npm start
+
+.PHONY: Test
+test:
```
joefong-nhs, line 72: would it be good to name this one as test-app or test-fe?  Because the same Makefile is shared across app(frontend) and backend(lambdas)

PR #23, opened 2023-08-30
Prmdr subtask 113

RachelHowellNHS:


app/src/helpers/test/testBuilders.ts
```diff
@@ -0,0 +1,58 @@
...
+    const mockDocument: UploadDocument = {
+        file,
+        state: uploadStatus ?? documentUploadStates.SUCCEEDED,
+        progress: 0,
+        id: Math.floor(Math.random() * 1000000).toString(),
+    };
+    return mockDocument;
+};
+
+const buildSearchResult = (searchResultOverride: any) => {
```
rio-k9, line 47: Although it wasn't part of the ticket can we quickly make a type for a search result here, using the example data types below. Similarly we then want here `Partial<SearchResult>` as the override type. and `SearchResult` as the function return type

rio-k9 (changes requested):

PR #39, opened 2023-09-07
PRMDR:90 - Migrate Document Search views and update lambda response

abbas-khan10:
with-results
without-results


app/src/helpers/requests/documentSearchResults.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const { data }: GetDocumentSearchResultsResponse = await axios.get(gatewayUrl, {
+        headers: {
+            'Content-Type': 'application/json',
+        },
+        params: {
+            patientId: nhsNumber,
+        },
+    });
+    return data;
```
RioKnightleyNHS, line 25: Can we wrap this entire method in a Try Catch too, but in the catch just throw the error upwards where we also catch in the page. It's just good practise for method re-usability

app/src/pages/documentSearchResultsPage/DocumentSearchResultsPage.tsx
```diff
@@ -0,0 +1,142 @@
...
+        }, 2000);
+    };
+
+    return (
+        <>
+            <h1>Download electronic health records and attachments</h1>
+
+            {(submissionState === SUBMISSION_STATE.FAILED ||
+                downloadState === SUBMISSION_STATE.FAILED) && <ServiceError />}
+
```
RioKnightleyNHS, line 72: Sometimes when there are lots of boolean operators controlling the render of page parts, I make a list of constants at the top of the page something like:

const searchIsPending = submissionState === SUBMISSION_STATE
const searchHasFailed = submissionState === SUBMISSION_STATE.FAILED
const searchHasSucceeded = submissionState === SUBMISSION_STATE.SUCCESS
const searchIsFinished = searchHasFailed || searchHasSucceeded
const noResults = searchResults.length === 0

RioKnightleyNHS: It makes it much easier to digest the render conditions when we use these values in the render operators then. But i'd say this is optional more than required. Depends what you find easier to read

lambdas/models/config.py
```diff
@@ -0,0 +1,9 @@
+from pydantic import ConfigDict
+
+
+def to_camel(string: str) -> str:
+    string_split = string.split("_")
+    return string_split[0] + "".join(word.capitalize() for word in string_split[1:])
+
```
RioKnightleyNHS, line 7: Consider the location of util functions. Optional

PR #42, opened 2023-09-11
Prmdr 163 migrate authorisation handler

joefong-nhs:
- Add an authoriser lambda which verify and decode a JWT authorizationToken, and respond with IAM policy which allow or deny user access to other APIs
- Add unit test for above lambda
- Change `make zip` command to pack with `--platform manylinux2014_x86_64` flag, as python's cryptography package seems to require some platform specific native library

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ env:
 zip:
 	rm -rf ./lambdas/package_$(lambda_name) || true 
 	mkdir ./lambdas/package_$(lambda_name)
-	./lambdas/venv/bin/pip3 install --cache-dir .pip_cache -r lambdas/requirements.txt -t ./lambdas/package_$(lambda_name)
+	./lambdas/venv/bin/pip3 install --platform manylinux2014_x86_64 --only-binary=:all: --implementation cp  -r lambdas/requirements.txt -t ./lambdas/package_$(lambda_name)
 	mkdir ./lambdas/package_$(lambda_name)/handlers
 	cp -r lambdas/handlers/$(lambda_name).py lambdas/package_$(lambda_name)/handlers
 	cp -r lambdas/utils lambdas/package_$(lambda_name)
--- a/lambdas/handlers/authoriser_handler.py
+++ b/lambdas/handlers/authoriser_handler.py
@@ -2,9 +2,12 @@
 This code has been modified from AWS blueprint. Below is the original license:
 
 Copyright 2015-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
-Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the License. A copy of the License is located at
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -46,7 +46,7 @@ env:
 zip:
 	rm -rf ./lambdas/package_$(lambda_name) || true 
 	mkdir ./lambdas/package_$(lambda_name)
-	./lambdas/venv/bin/pip3 install --cache-dir .pip_cache -r lambdas/requirements.txt -t ./lambdas/package_$(lambda_name)
+	./lambdas/venv/bin/pip3 install --platform manylinux2014_x86_64 --only-binary=:all: --implementation cp  -r lambdas/requirements.txt -t ./lambdas/package_$(lambda_name)
```
SRAlexander, line 49: We should give this a good test, but it sounds like a reasonable adjustment.

lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,221 @@
...
+        logger.info(f"decoded JWT: {decoded}")
+    except botocore.exceptions.ClientError as e:
+        logger.error(e)
+    except jwt.PyJWTError as e:
+        logger.error(f"error while decoding JWT: {e}")
+    except KeyError as e:
+        logger.error(e)
+
+    principalId = ""
+    tmp = event["methodArn"].split(":")
```
SRAlexander, line 53: can we rename this to something more meaningful?

lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,221 @@
...
+import botocore
+import jwt
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+
+def lambda_handler(event, context):
+    user = "Unauthorised"
+    ssm_public_key_parameter_name = os.environ["SSM_PARAM_JWT_TOKEN_PUBLIC_KEY"]
```
SRAlexander, line 27: How are we planning to set the os.env on the aws lambda instance?

SRAlexander (changes requested):
Looks good, we will have to look at the deployment of lambda but we will have to wait until we can merge the infrastructure side.

lambdas/handlers/authoriser_handler.py
joefong-nhs: Thanks!  Addressed in this commit: https://github.com/nhsconnect/national-document-repository/pull/42/commits/c0192931217fdc6f88a751322e92b745595cfa75

lambdas/handlers/authoriser_handler.py
joefong-nhs: In a branch for subtask 165, we added this env var to the lambda by terraform : https://github.com/nhsconnect/national-document-repository-infrastructure/blob/prmdr_165_subtask/infrastructure/lambda-authoriser.tf

PR #48, opened 2023-09-18
Prmdr 167 migrate token handler

joefong-nhs:
- Add an OIDC Service which interact with CIS2
- Add an ODS API Service which retrieve user org info from ODS ORD API
- Add a token handler lambda, which verify user with CIS2 and respond with user org info, create login session and NDR's auth token
- Amend authoriser handler to adapt to new token structure, and add logic to verify user's login session exist in dynamo

**Note:** The WIP branch of PRMDR-165 is changing the API of our dynamo db service. To avoid merge conflict, the lambda in this PR will directly use boto3 dynamo resource for now. To be refactored after merged with PRMDR-165.


lambdas/handlers/authoriser_handler.py
```diff
@@ -33,20 +37,26 @@ def lambda_handler(event, context):
...
-        # For now we just assume the user authenticate with a signed token in this form:
-        # { "user_role": "Unauthorised"  or  "GP"  or  "PCSE" }
-        # The token is going to be issued by the lambda of ticket PRMDR-167
-
         decoded = jwt.decode(
             event["authorizationToken"], public_key, algorithms=["RS256"]
         )
-        user = decoded["user_role"]
-        logger.info(f"decoded JWT: {decoded}")
+        logger.debug(f"decoded JWT auth token: {decoded}")
```
SRAlexander, line 43: Are we planning on removing this?

lambdas/handlers/authoriser_handler.py
```diff
@@ -72,6 +83,35 @@ def lambda_handler(event, context):
     return auth_response
 
 
+def find_login_session(ndr_session_id):
+    logger.debug("Retrieving session for session ID ending in: " + ndr_session_id[-4:])
```
SRAlexander, line 87: -4 does not mean much; should this be created as a var so that we know what that array position links to?

SRAlexander (changes requested):

lambdas/handlers/authoriser_handler.py
SRAlexander: Seems a bit naughty to log all JWT tokens, maybe both should be for debug mode.

SRAlexander:
Minor recommendations but looking good :0

SRAlexander:
Minor recommendations but looking good :)

SRAlexander:
Minor recommendations but looking good :)

lambdas/handlers/authoriser_handler.py
joefong-nhs: Agree. Logging this token was just for dev purpose, so now this logging is removed.

lambdas/handlers/authoriser_handler.py
joefong-nhs: Thanks.  This is for extracting the last 4 chars in session ID for logging purpose.
It is now extracted as a tiny helper func `redact_id` for readability

lambdas/handlers/token_handler.py
```diff
@@ -0,0 +1,137 @@
...
+    # issue Authorisation token
+    ssm_client = boto3.client("ssm")
+    logger.info("starting ssm request to retrieve NDR private key")
+    ssm_response = ssm_client.get_parameter(
+        Name="jwt_token_private_key", WithDecryption=True
+    )
+    logger.info("ending ssm request")
+    private_key = ssm_response["Parameter"]["Value"]
+    ndr_token_content = {
+        "exp": min(time.time() + 60 * 30, id_token_claim_set.exp),
```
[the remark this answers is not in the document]
joefong-nhs: Agree. Now changed it like this for clarity:

```python
    thirty_minutes_later = time.time() + 60 * 30
    ndr_token_expiry_time = min(thirty_minutes_later, id_token_claim_set.exp)

    ndr_token_content = {
        "exp": ndr_token_expiry_time,
        ...
```

PR #49, opened 2023-09-18
Prmdr 165 subtask

joefong-nhs:


lambdas/tests/unit/services/test_dynamo_query_service.py
```diff
@@ -91,19 +93,29 @@ def test_lambda_handler_returns_items_from_dynamo(mock_dynamo_table, mock_boto3_
 def test_error_raised_when_no_fields_requested(mock_dynamo_table, mock_boto3_dynamo):
     with patch.object(boto3, "resource", return_value=mock_boto3_dynamo):
         mock_boto3_dynamo.Table.return_value = mock_dynamo_table
-        query_service = DynamoQueryService("test_table", "test_index")
+        query_service = DynamoDBService(
+            "test_table",
+        )
         with pytest.raises(InvalidResourceIdException):
-            query_service("NhsNumber", "0123456789", [])
+            query_service.query_service("test_index", "NhsNumber", "0123456789", [])
```
SRAlexander, line 100: Should this be called DynamoDbServiceMock to avoid calling query_service.query_service?

lambdas/handlers/login_redirect_handler.py
```diff
@@ -0,0 +1,68 @@
...
+    dynamodb_name = os.environ["AUTH_DYNAMODB_NAME"]
+    dynamodb_service = DynamoDBService(dynamodb_name)
+    ttl = round(time.time()) + 60 * 10
+    item = {"State": state, "TimeToExist": ttl}
+    dynamodb_service.post_item_service(item=item)
+
+
+def get_ssm_parameters():
+    ssm_parameters_names = ["OIDC_AUTHORISE_URL", "OIDC_CLIENT_ID"]
+    ssm_client = boto3.client("ssm", region_name="eu-west-2")
```
SRAlexander, line 67: Should region be an env var so that it's not hard coded?

lambdas/handlers/login_redirect_handler.py
```diff
@@ -0,0 +1,68 @@
...
+        ).create_api_gateway_response()
+    return ApiGatewayResponse(302, "", "GET").create_api_gateway_response(
+        headers=location_header
+    )
+
+
+def save_state_in_dynamo_db(state):
+    dynamodb_name = os.environ["AUTH_DYNAMODB_NAME"]
+    dynamodb_service = DynamoDBService(dynamodb_name)
+    ttl = round(time.time()) + 60 * 10
```
SRAlexander, line 60: Time as a var e.g. TenMinuteTimespan = time.time() + 60 * 10;

SRAlexander (changes requested):

lambdas/tests/unit/services/test_dynamo_query_service.py
joefong-nhs: Thanks. We renamed the mock dynamodb service to `db_service`, as Noga did in some other tests

lambdas/handlers/login_redirect_handler.py
joefong-nhs: For the region_name, we replaced the hardcoding with a tiny util function `get_aws_region` which read from env var  AWS_REGION, and if not provided will return the default "eu-west-2"

lambdas/handlers/login_redirect_handler.py
joefong-nhs: Thanks! We replaced 60 * 10 to variable "ten_minutes" for clarity

PR #51, opened 2023-09-21
Prmdr 116 2

SRAlexander:


lambdas/handlers/create_document_reference_handler.py
```diff
@@ -19,23 +20,42 @@
...
+        response = ApiGatewayResponse(400, "An error occured processing the required document type", "POST").create_api_gateway_response()
+        return response
+    
+    # We cannot get here unless a valid document type has been provided, therefore we can default to ARF
     try:
         s3_bucket_name = os.environ["DOCUMENT_STORE_BUCKET_NAME"]
         dynamo_table = os.environ["DOCUMENT_STORE_DYNAMODB_NAME"]
 
-        body = json.loads(event["body"])
+        if (document_type == SupportedDocumentTypes.LG.name):
```
AlexHerbertNHS, line 39: Might be more readable if this was something like
```
if document_type == "LG":
    s3_bucket_name = os.environ["LG_BUCKET"]
else:
    s3_bucket_name = os.environ["ARF_BUCKET"]

```
should work just fine, as Python is a bit looser in the scoping than Java so you don't need to declare the variable before the if statement

lambdas/handlers/create_document_reference_handler.py
```diff
@@ -19,23 +20,42 @@
 def lambda_handler(event, context):
     logger.info("Starting document reference creation process")
 
+    try:
+        document_type = (event["queryStringParameters"]["documentType"]).upper()
+        if document_type in SupportedDocumentTypes.list_names():
```
AlexHerbertNHS, line 25: I personally find it useful to give my enums a method where you can pass in a value and they return the appropriate Enum for that value, or None if there's not one. Might make it a bit more readable if you do:

```
document_type = SupportedDocumentTypes.from_value((event["queryStringParameters"]["documentType"]).upper())
if document_type is None:
    raise Exception
logger.info("Provided document is supported")
```

Then you'll have doc type as the enum itself so you don't have to check against SupportedDocumentTypes.LG.name below, you can drop the ".name"

lambdas/handlers/create_document_reference_handler.py
```diff
@@ -19,23 +20,42 @@
...
+            logger.info("Provided document is supported")
+        else:
+            raise Exception("Provided document type is not supported")
```
AlexHerbertNHS, line 28: In Java, it's good practice to never raise / throw base Exception class. Not sure about Python, but this gives me the willies.

lambdas/tests/unit/handlers/test_create_document_reference_handler.py
```diff
@@ -30,18 +35,82 @@
...
+def arf_type_event():
     api_gateway_proxy_event = {
+        "queryStringParameters": {"documentType": "arf"},
         "body": '{"subject": {"identifier": {"value": "test"}}, '
         '"content": [{"attachment": {"contentType": "test"}}], '
         '"description": "test"}'
     }
     return api_gateway_proxy_event
 
+@pytest.fixture
```
AlexHerbertNHS, line 55: I think abbas wanted fixtures that are used in multiple places to be moved to conftest.py, would this be a candidate for that?

lambdas/tests/unit/services/test_dynamo_service_get_table.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@pytest.fixture
+def mock_boto3_dynamo():
+    return MagicMock()
+
+
+@pytest.fixture
+def mock_dynamo_table():
+    return MagicMock()
```
AlexHerbertNHS, line 17: I also think that Abbas was concerned about us using overlapping test libraries and we were to avoid MagicMock in favour of Mocker, but I'm not sure on the status of that. I've always had to use with patch.object myself, not really a change just one to be aware of

AlexHerbertNHS (commented):
just some minor things, happy to discuss if you'd prefer.

lambdas/services/dynamo_service.py
```diff
@@ -10,32 +10,36 @@
 
 
 class DynamoDBService:
-    def __init__(self, table_name):
+
+    def get_table(self, table_name):
```
AlexHerbertNHS, line 14: If we're removing the init method, then there will be no point in this being a class anymore. We should de-class this and just have methods to do the needful, I think. The only references to self are self.get_table, and that in itself is a static method.

lambdas/handlers/create_document_reference_handler.py
```diff
@@ -20,27 +20,25 @@
...
-        document_type = (event["queryStringParameters"]["documentType"]).upper()
-        if document_type in SupportedDocumentTypes.list_names():
-            logger.info("Provided document is supported")
-        else:
-            raise Exception("Provided document type is not supported")   
-    except Exception as e:
+    document_type_string = (event["queryStringParameters"]["documentType"]).upper()
+    document_type = SupportedDocumentTypes.get_from_field_name(document_type_string);
+    if document_type is None:
         logger.error(e)
```
AlexHerbertNHS, line 26: There's no error event called e anymore. Please remove this and log out something otherwise we will get errors here

AlexHerbertNHS (changes requested):
One small change please

AlexHerbertNHS (approved):
LGTM 👍

PR #57, opened 2023-09-26
Prmdr 161

joefong-nhs:
Story branch PRMDR-161  `ACC1.1 | Migrate CIS2 Lambdas`  to be merged into main


lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,268 @@
...
+This code has been modified from AWS blueprint. Below is the original license:
+
+Copyright 2015-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Licensed under the Apache License, Version 2.0 (the "License").
+You may not use this file except in compliance with the License. A copy of the License is located at
+     http://aws.amazon.com/apache2.0/
+or in the "license" file accompanying this file.
+This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and limitations under the License.
+"""
```
abbas-khan10, line 11: Do we need to keep this comment in?

lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,268 @@
...
+
+from utils.get_aws_region import get_aws_region
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+
+def lambda_handler(event, context):
+    user_roles = []
+    ssm_public_key_parameter_name = os.environ["SSM_PARAM_JWT_TOKEN_PUBLIC_KEY"]
```
abbas-khan10, line 34: There's no check to see if this env variable exists before it's use, we use `os.getenv` in some lambdas which it would raise a key error however since we're using `environ` here maybe a check to see if `ssm_public_key_parameter_name` has a value and return 500 if nothing

lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,268 @@
...
+
+    try:
+        client = boto3.client("ssm", region_name=get_aws_region())
```
abbas-khan10, line 37: Might be worth noting we'll be using SSM in other places (i.e PDS integration), so could think about extracting and generalising SSM code into a service as we've done with other AWS clients

lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,268 @@
...
+def validate_login_session(current_session, ndr_session_id):
+    expiry_time = current_session["TimeToExist"]
+    time_now = time.time()
+    if expiry_time <= time_now:
+        raise AuthorisationException(
+            f"The session is already expired for session ID ending in: {redact_id(ndr_session_id)}"
+        )
+
+
+class HttpVerb:
```
abbas-khan10, line 123: This file feels like it's getting pretty big with lots of components for just a lambda handler, maybe could extract these out into utils or models to clean up the file slightly?

lambdas/handlers/login_redirect_handler.py
```diff
@@ -0,0 +1,71 @@
...
+        return ApiGatewayResponse(
+            500, "Server error", "GET"
+        ).create_api_gateway_response()
+    return ApiGatewayResponse(302, "", "GET").create_api_gateway_response(
+        headers=location_header
+    )
+
+
+def save_state_in_dynamo_db(state):
+    dynamodb_name = os.environ["AUTH_DYNAMODB_NAME"]
```
abbas-khan10, line 60: May need a check on this env variable

lambdas/services/dynamo_service.py
```diff
@@ -58,6 +58,20 @@ def query_service(
             logger.error(e)
             raise e
 
+    def simple_query(self, table_name: str, key_condition_expression):
```
abbas-khan10, line 61: We should maybe think about the naming of some of these functions, to me they're not very clear. What is a simple query? the `key_condition_expression` is missing a type so it may make it hard for someone to know how to implement it

lambdas/services/dynamo_service.py
```diff
@@ -82,6 +96,18 @@ def update_item_service(self, table_name, key, update_expression, expression_att
             logger.error(e)
             raise e
 
+    def delete_item_service(self, table_name: str, key: dict):
```
abbas-khan10, line 99: Same here, the function belongs to a service and so is implemented `DynamoDBService.delete_item_service` doesn't really make sense to have `service` mentioned twice? Could we maybe rename these `service` functions to just be what the function is doing?

lambdas/handlers/authoriser_handler.py
NogaNHS: Yes, legally we need this in.

lambdas/requirements.txt
```diff
@@ -7,7 +7,10 @@ cryptography==41.0.2
 idna==3.4
 Jinja2==3.1.2
 jmespath==1.0.1
+oauthlib==3.2.2
```
NogaNHS, line 10: duplication with line 12

lambdas/services/oidc_service.py
```diff
@@ -0,0 +1,145 @@
...
+        "verify_nbf": True,
+        "verify_iat": True,
+        "verify_aud": True,
+        "verify_iss": True,
+    }
+
+    def __init__(self):
+        oidc_parameters = self.fetch_oidc_parameters()
+
+        # logger.info(f"Initialising OIDC service with parameter: {oidc_parameters}")
```
abbas-khan10, line 31: Using this log?

lambdas/tests/unit/utils/test_get_aws_region.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+from utils.get_aws_region import get_aws_region
+
+
+def test_get_aws_region_return_aws_region_in_env_var(mocker):
```
abbas-khan10, line 6: these environment variables and test variables have been set as a fixture in `tests/unit/conftest.py` we can re use these

lambdas/tests/unit/services/test_oidc_service.py
```diff
@@ -0,0 +1,238 @@
...
+            "error": "invalid_token",
+        },
+    )
+    mocker.patch("requests.get", return_value=mock_response)
+
+    with pytest.raises(AuthorisationException):
+        oidc_service.fetch_user_org_codes(mock_token)
+
+
+# Below are tests around the validate_and_decode_token method. Involve mocking the CIS2 signing key pairs
```
abbas-khan10, line 166: I'd personally avoid comments like this, code should generally speak for itself rather than littering the code base with little comments like this unless it's obscure?

lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,268 @@
...
+        decoded = jwt.decode(
+            event["authorizationToken"], public_key, algorithms=["RS256"]
+        )
+
+        ndr_session_id = decoded["ndr_session_id"]
+
+        current_session = find_login_session(ndr_session_id)
+        validate_login_session(current_session, ndr_session_id)
+
+        # if user has a valid session, assign their role
```
abbas-khan10, line 52: is this comment necessary?

lambdas/utils/get_aws_region.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+
+DEFAULT_REGION = "eu-west-2"
+
+
+def get_aws_region() -> str:
+    return os.environ.get("AWS_REGION", DEFAULT_REGION)
```
abbas-khan10, line 8: This function is not necessary, it's covered by the `set_env` function in the global conftest.py

lambdas/handlers/authoriser_handler.py
```diff
@@ -0,0 +1,268 @@
...
+        user_roles = [org["role"] for org in decoded["organisations"]]
+
+    except AuthorisationException as e:
+        logger.error(e)
+    except botocore.exceptions.ClientError as e:
+        logger.error(e)
+    except jwt.PyJWTError as e:
+        logger.error(f"error while decoding JWT: {e}")
+    except (KeyError, IndexError) as e:
+        logger.error(e)
```
abbas-khan10, line 62: What's happening in these except blocks? Once the error is logged the code will continue on and return the auth response. Is this what we want?

lambdas/handlers/login_redirect_handler.py
```diff
@@ -0,0 +1,71 @@
...
+        oidc_parameters = {
+            parameter["Name"]: parameter["Value"]
+            for parameter in ssm_response["Parameters"]
+        }
+
+        oidc_client = web_application_client_class(
+            client_id=oidc_parameters["OIDC_CLIENT_ID"],
+        )
+
+        url, headers, body = oidc_client.prepare_authorization_request(
```
AlexHerbertNHS, line 34: Headers and Body don't seem to be used

lambdas/services/oidc_service.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self._client_id = oidc_parameters["OIDC_CLIENT_ID"]
+        self._client_secret = oidc_parameters["OIDC_CLIENT_SECRET"]
+        self._oidc_issuer_url = oidc_parameters["OIDC_ISSUER_URL"]
+        self._oidc_token_url = oidc_parameters["OIDC_TOKEN_URL"]
+        self._oidc_userinfo_url = oidc_parameters["OIDC_USER_INFO_URL"]
+        self._oidc_callback_uri = oidc_parameters["OIDC_CALLBACK_URL"]
+        self._oidc_jwks_url = oidc_parameters["OIDC_JWKS_URL"]
+        self.scope = "openid profile nationalrbacaccess associatedorgs"
```
AlexHerbertNHS, line 40: self.scope doesn't appear to be used

lambdas/tests/unit/helpers/data/ods/__init__.py
AlexHerbertNHS, line 1: Empty file - is this needed?

lambdas/tests/unit/helpers/data/ods/ods_org_response.py
AlexHerbertNHS, line 1: Empty file - is this needed?

lambdas/utils/get_aws_region.py
AlexHerbertNHS, line 8: Kinda seems like a lot to me to have this one file to run one line of code.

lambdas/services/oidc_service.py
joefong-nhs: Thanks! removed this log.

lambdas/handlers/authoriser_handler.py
joefong-nhs: At the auth response part, if the user doesn't have a valid role, we responds with Deny All by default.
I have edited these except block and immediately respond with a Deny All here

lambdas/handlers/authoriser_handler.py
joefong-nhs: We kept those class in the lambda handler because those are from AWS blueprint so we tried to keep them in one place, but I am happy to move the classes to a separate file.

lambdas/tests/unit/helpers/data/ods/ods_org_response.py
joefong-nhs: Thanks for spotting that! Binned that empty file

lambdas/handlers/authoriser_handler.py
joefong-nhs: Probably not. Comment removed :)

lambdas/tests/unit/services/test_oidc_service.py
joefong-nhs: Thanks! Removed the redundant comment.

lambdas/handlers/token_handler.py
```diff
@@ -0,0 +1,146 @@
...
+
+    return session_id
+
+
+def issue_auth_token(
+    session_id: str,
+    id_token_claim_set: IdTokenClaimSet,
+    permitted_orgs_and_roles: list[dict],
+) -> str:
+    ssm_client = boto3.client("ssm")
```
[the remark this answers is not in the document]
richbyrne-nhs: Add this to a future ticket - there'll be a few services we can reuse and we can capture that under a future story

lambdas/handlers/authoriser_handler.py
richbyrne-nhs: Add to a future ticket - it'll be picked up as code smells and we can make sure we refactor when everything is in place

lambdas/handlers/login_redirect_handler.py
NogaNHS: Will be done as part of the decorator work

lambdas/tests/unit/helpers/data/ods/__init__.py
NogaNHS: Yes

lambdas/services/oidc_service.py
NogaNHS: Deleted

lambdas/services/dynamo_service.py
NogaNHS: Done

lambdas/tests/unit/utils/test_get_aws_region.py
NogaNHS: Done

lambdas/utils/get_aws_region.py
NogaNHS: Done

lambdas/handlers/login_redirect_handler.py
NogaNHS: prepare_authorization_request returns a tuple so we've to assign it.

PR #59, opened 2023-09-26
PRMDR_99: Document Search now uses an array of tables rather than individual tables

SRAlexander:


lambdas/tests/unit/handlers/test_document_reference_search_handler.py
```diff
@@ -118,3 +118,12 @@ def test_lambda_handler_when_id_not_supplied_returns_400(
     ).create_api_gateway_response()
     actual = lambda_handler(missing_id_event, context)
     assert expected == actual
+
+def test_lambda_handler_when_dynamo_tables_env_variable_not_supplied_then_return_400_response(
+valid_id_event, context
+):
+    expected = ApiGatewayResponse(
+        400, "An error occurred due to missing key: 'DYNAMODB_TABLE_LIST'", "GET"
+    ).create_api_gateway_response()
```
joefong-nhs, line 127: Just thinking should we be responding with 400 or 5xx for this kind of error?
Because usually missing an env var probably means something wrong in our config, which the user has no way to fix. 
Responding 400 with missing key: 'DYNAMODB_TABLE_LIST' seems to imply that the request should have a param / header of DYNAMODB_TABLE_LIST, which isn't the case.

abbas-khan10: I agree, however this may require a change the way the code handles env variables and raise a custom exception if any are missing, as right now it catches KeyError (also used for checking for the existence of query params for nhs number etc) which should be a 400 in some cases but then 5xx for env variables

abbas-khan10 (approved):

PR #61, opened 2023-09-28
Task lambda action simplfication

SRAlexander:


SRAlexander:
Force merging new actions into the main branch while leaving the existing actions as they are, we cannot test the simplified actions untill they exist on main.

PR #63, opened 2023-09-28
Action optimisations

SRAlexander:


.github/workflows/new_lambdas-deploy-to-perf.yml
```diff
@@ -17,28 +17,24 @@ permissions:
...
-  SANDBOX: ${{ inputs.sandbox }}
-
 jobs:
   run_tests:
     uses: ./.github/workflows/new_base-lambdas-reusable-test.yml
     with:
-      environment: ${{ env.ENVIRONMENT }}
-      python-version: ${{ env.PYTHON_VERSION }}
-      build-branch: ${{ env.BUILD_BRANCH }}
+      environment: perf
```
abbas-khan10, line 24: Is there a reason why we're hard coding this and subsequent environments instead of using `inputs.environment`

SRAlexander: Cheers for taking a quick look, but yup this one is specific to perf so we don't want to open this one up to additional environments that can be input at runtime.

PR #64, opened 2023-09-28
[PRMDR-99] - Fix error when parsing list of dynamo tables from environment variables

abbas-khan10:


RachelHowellNHS (approved):
Code LGTM and tested on ndr-test - upload journey now works :)

PR #72, opened 2023-10-04
PRMDR-35 Auth CIS2 (no tests)

RioKnightleyNHS:


app/src/helpers/test/testBuilders.ts
```diff
@@ -6,6 +6,21 @@ import {
...
+        organisations: [
+            {
+                org_name: 'PORTWAY LIFESTYLE CENTRE',
+                ods_code: 'A470',
+                role: 'DEV',
+            },
+        ],
+        authorisation_token: '111xxx222',
+    };
+    return auth;
```
joefong-nhs, line 22: I think probably a {...userAuthOverride} is missing here?  If it is just the test is WIP then no probs.

PR #73, opened 2023-10-04
PRMDR-270 & PRMDR-268

joefong-nhs:


app/src/helpers/requests/lloydGeorgeSearchResult.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type LloydGeorgeStitchResult = {
+    number_of_files: number;
+    total_file_size_in_byte: number;
+    last_updated: string;
+    presign_url: string;
+};
+
+async function getLloydGeorgeRecord({
+    nhsNumber,
```
RioKnightleyNHS, line 16: With these helper requests generally the file name should match the name of the exported function.

Rename file to getLloydGeorgeRecord or vice versa

app/src/helpers/requests/lloydGeorgeSearchResult.ts
```diff
@@ -0,0 +1,32 @@
...
+    baseUrl,
+}: Args): Promise<LloydGeorgeStitchResult> {
+    const gatewayUrl = baseUrl + '/LloydGeorgeStitch';
+
+    const { data } = await axios.get(gatewayUrl, {
+        headers: {
+            'Content-Type': 'application/json',
```
RioKnightleyNHS, line 23: This will need to use the new "useBaseAPIHeaders" which will contain the auth data, it's merged into main now so rebase and add that hook in similar to how baseUrl hook works. Find another request example :)

app/src/pages/lloydGeorgeRecordPage/LloydGeorgeRecordPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+function LloydGeorgeRecordPage() {
+    const [patientDetails] = usePatientDetailsContext();
+    const [downloadStage, setDownloadStage] = useState(DOWNLOAD_STAGE.INITIAL);
+    const [numberOfFiles, setNumberOfFiles] = useState(0);
+    const [totalFileSizeInByte, setTotalFileSizeInByte] = useState(0);
+    const [lastUpdated, setLastUpdated] = useState('');
+    const [lloydGeorgeUrl, setLloydGeorgeUrl] = useState('');
+    const navigate = useNavigate();
+    const baseUrl = useBaseAPIUrl();
+
```
RioKnightleyNHS, line 23: Add useBaseAPIHeaders here

app/src/pages/lloydGeorgeRecordPage/LloydGeorgeRecordPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+                {`${patientDetails?.givenName} ${patientDetails?.familyName}`}
+            </p>
+            <p style={{ fontSize: '16px', marginBottom: 5 }}>NHS number: {nhsNumber}</p>
+            <p style={{ fontSize: '16px' }}>Date of birth: {dob}</p>
+        </>
+    );
+
+    useEffect(() => {
+        if (!patientDetails) {
+            navigate(routes.HOME);
```
RioKnightleyNHS, line 47: This block of code for checking if a patient exists then returning home, no longer necessary on pages as the 'PatientGuard' component on the Router handles this and takes us to 'Unauthorised'

RioKnightleyNHS: Just have everything in the else statement

app/src/pages/lloydGeorgeRecordPage/LloydGeorgeRecordPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+                        setTotalFileSizeInByte(total_file_size_in_byte);
+                    }
+                    setDownloadStage(DOWNLOAD_STAGE.SUCCEEDED);
+                } catch (e) {
+                    setDownloadStage(DOWNLOAD_STAGE.FAILED);
+                }
+            };
+            void search();
+        }
+    }, [patientDetails, baseUrl, navigate]);
```
RioKnightleyNHS, line 72: Does this cause a memory leak? I had to fix the last document search code that looked similar to this search function. 
See DocumentSearchResults for example with using a ref.mounted and toggling it so we know this search just runs once.

app/src/pages/lloydGeorgeRecordPage/LloydGeorgeRecordPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+            return <>No documents are available</>;
+        } else {
+            return <>Loading...</>;
+        }
+    };
+
+    return (
+        <>
+            <>{patientInfo}</>
+            <Card style={{ marginBottom: 0 }}>
```
RioKnightleyNHS, line 96: Probably fine for now to write style like this. But there needs to be a team discussion on whether we discontinue SCSS files in favour of JSX style. It's generally bad to mix

app/src/types/generic/routes.ts
```diff
@@ -11,6 +11,8 @@ export enum routes {
     DOWNLOAD_DOCUMENTS = '/search/results',
     DELETE_DOCUMENTS = '/search/results/delete',
 
+    LLOYD_GEORGE = '/lloyd-george-record',
```
RioKnightleyNHS, line 14: Has this URL been agreed with ? Seems a bit odd to not have any more sub-directories in the path as you will have navigated through some pages by this

RioKnightleyNHS (changes requested):

SRAlexander (approved):

PR #82, opened 2023-10-10
Prmdr 336 subtask

joefong-nhs:


lambdas/handlers/bulk_upload_metadata_handler.py
```diff
@@ -0,0 +1,91 @@
...
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+
+def lambda_handler(_event, _context):
+    try:
+        logger.info("Starting metadata reading process")
+
+        staging_bucket_name = os.environ["STAGING_STORE_BUCKET_NAME"]
```
AlexHerbertNHS, line 22: Should these not be checked by the ensure_env_var decorator that @joefong-nhs has written previously?

lambdas/handlers/bulk_upload_metadata_handler.py
```diff
@@ -0,0 +1,91 @@
...
+        logger.info("Failed to parse metadata.csv")
+        logger.error(str(e))
+    except KeyError as e:
+        logger.info("Failed due to missing key")
+        logger.error(str(e))
+    except ClientError as e:
+        logger.error(str(e))
+
+
+def download_metadata_from_s3(staging_bucket_name: str, metadata_filename: str):
```
AlexHerbertNHS, line 47: Could this method be moved to an inherited version of the S3service to keep this file a little cleaner?

lambdas/handlers/bulk_upload_metadata_handler.py
```diff
@@ -0,0 +1,91 @@
...
+        for nhs_number in patients
+    ]
+
+
+def send_metadata_to_sqs(
+    staging_metadata_list: list[StagingMetadata], metadata_queue_url: str
+) -> None:
+    sqs_service = SQSService()
+
+    for staging_metadata in staging_metadata_list:
```
AlexHerbertNHS, line 83: Could this method be moved to an inherited version of the SQS service to keep this file a little cleaner?

lambdas/tests/unit/handlers/test_bulk_upload_metadata_handler.py
```diff
@@ -0,0 +1,166 @@
...
+    assert actual == expected
+
+
+def test_csv_to_staging_metadata():
+    actual = csv_to_staging_metadata(MOCK_METADATA_CSV)
+    expected = EXPECTED_PARSED_METADATA
+    assert actual == expected
+
+
+def test_send_metadata_to_sqs(mock_sqs_service):
```
AlexHerbertNHS, line 126: Only the happy path is tested for this function. I'd like to see a test for what would happen if it gets some unexpected data plugged in, and see it raises the expected error

lambdas/tests/unit/handlers/test_bulk_upload_metadata_handler.py
```diff
@@ -0,0 +1,166 @@
...
+
+    mock_s3_service.download_file.assert_called_with(
+        s3_bucket_name=MOCK_LG_STAGING_STORE_BUCKET,
+        file_key=METADATA_FILENAME,
+        download_path=f"{MOCK_TEMP_FOLDER}/{METADATA_FILENAME}",
+    )
+    assert actual == expected
+
+
+def test_csv_to_staging_metadata():
```
AlexHerbertNHS, line 120: Only the happy path is tested for this function. I'd like to see a test for what would happen if it gets some unexpected data plugged in, and see it raises the expected error

lambdas/tests/unit/handlers/test_bulk_upload_metadata_handler.py
```diff
@@ -0,0 +1,166 @@
...
+        " The specified queue does not exist"
+    )
+
+    lambda_handler(None, None)
+
+    assert caplog.records[-1].message == expected_err_msg
+    assert caplog.records[-1].levelname == "ERROR"
+
+
+def test_download_metadata_from_s3(mock_s3_service, mock_tempfile):
```
AlexHerbertNHS, line 105: Only the happy path is tested for this function. I'd like to see a test for what would happen if it gets some unexpected data plugged in, and see it raises the expected error

AlexHerbertNHS (approved):
Generally good, a few changed tho. Happy to discuss

AlexHerbertNHS (changes requested):
Accidentally clicked Approve instead of Request Changes bvefore - D'Oh!

lambdas/handlers/bulk_upload_metadata_handler.py
joefong-nhs: We thought of using that, but that decorator will respond with 500 status, and this lambda ins't going to be called from a web API, so finally we choose to just wrap this in a try block and catch KeyErrors

lambdas/tests/unit/handlers/test_bulk_upload_metadata_handler.py
joefong-nhs: Thanks. Added test for unhappy path (when the data in csv file is invalid)

lambdas/tests/unit/handlers/test_bulk_upload_metadata_handler.py
joefong-nhs: Added test for unhappy path (fail to download from S3)

lambdas/tests/unit/handlers/test_bulk_upload_metadata_handler.py
joefong-nhs: This method is not responsible to validate the input data, so as long as the input is of correct type (class StagingMetadata) then it will send the message to sqs no matter how the data content is.
Now I added a test for a possible unhappy path scenario (if failed to send msg to sqs)

lambdas/handlers/bulk_upload_metadata_handler.py
joefong-nhs: For the suggestion of inherited versions of AWS related classes, I am a bit hesitant to do so.

If every lambda that interact with AWS create their own inherited class, then we will end up with n * m (number of lambda * number of AWS service we use) different classes, which could be difficult to maintain/understand.

In my conceptual model, AWS service classes are responsible to handle the communication with AWS, and should (hopefully) offer a simpler / more streamlined API than just using vanilla boto3. Lambda specific details (which bucket to use, what params to call with etc) are better to be handled by the lambda themselves. 
And if in case if we need to switch from AWS to say Azure in the future, hopefully we could do it by amending the m AWS service classes rather than having to amend (n * m) different flavours of AWS service classes.

AlexHerbertNHS: Hi Joe, thanks for your input on this. This one made me think real hard for a decent response

I guess my main gripe is that putting all of the logic in the lambda and then having a 'Service Class' with minimal code basically just gives you a Service class in name only which is just a wrapper around Boto3. Honestly, based on the way some of these are written I don't see why the lambdas themselves aren't just calling Boto3 themselves, as the 'Service' class is adding minimal value

This violates SOLID code standard - the Lambda/controller is not Single Purpose as it does everything itself, it's not Open/Close as you can't really extend the logic of the lambda to a sub-lambda that does similar things but in a different way, etc... - whereas if we extend the current service classes in the manner recommended by Scott (using this as an example, refactor this into a new class that extends S3Service), it keeps all the code to do this one thing in one place, making it more testable and more reusable if we need to do this outside of this lambda. Not perfectly SOLID, but at least it makes steps to get there.

I also feel that it violates the standard MVC pattern of controller -> service -> repository, I really feel that the Lambda (controller) should get the request, plug the data it gets into its constituent services where the bulk of the business logic lives, leaving the handler very lean and readable and having all the business logic in one place where it's reusable. 

I can kinda see what you mean with the n x m complexity, but I'd argue that each service class should fulfil a use case, so it should be n x use cases as opposed to lambdas x specific implementations. This also helps keep the units small and testable.

As for the change to Azure example - not only do I think this is highly unlikely that the NHS would suddenly pivot and require all of their AWS infrastructure to suddenly be moved to Azure. However, if they do do that for whatever reason, if the code sticks to the MVC implementation it should be trivial to refactor the base repository classes (currently S3Service, DynamoDBService etc...) to use Azure python libraries with no change to the classes that use it, and still produce the same result where possible, if the code was structured in the appropriate way.

However, I concede that the code is all over the place because we've just been running with it to go fast and not really considering best code practices. If you insist on this, I'l approve it for you, but I can't promise I won't refactor it all when we get time to discuss and implement standard coding practices in the app 😄 

Thanks again for this comment - really made me think about how we do things vs. how we should be doing them

joefong-nhs: Thanks Alex, really really appreciate how much thought you put into this issue.
Could we discuss this in a tech huddle this pm or tomorrow?  As I would like to hear the opinion of other team members on this. I am more than happy to adapt if other dev agree on the idea of inherited S3 service class per lambda.

AlexHerbertNHS: I'm happy to just approve this now and discuss it later when we all have time to think about coding standards - I don't want to block your current work for the sake of an in depth discussion

AlexHerbertNHS (approved):

PR #84, opened 2023-10-10
PRMDR-242 subtask BE LG Validator

NogaNHS:


lambdas/services/lloyd_george_validator.py
```diff
@@ -1,7 +1,44 @@
...
+class LGInvalidFilesException(Exception):
+    """One or more of the files are not valid"""
+    pass
+
+def validate_lg_file_type(file_type: str):
     if file_type != 'application/pdf':
         raise LGFileTypeException
+
+def validate_file_name(name: str):
+    lg_regex = r'[0-9]+of[0-9]+_Lloyd_George_Record_\[[A-Za-z]+\s[A-Za-z]+]_\[[0-9]{10}]_\[\d\d-\d\d-\d\d\d\d].pdf'
```
joefong-nhs, line 23: Can I suggest to change the regex part for patient name to something more flexible like `[a-zA-Z ]+` ?  Because I think we can't be sure that every full name is made up of exactly 2 words. For me my legal name (on passport) is 3 words

NogaNHS: Done

lambdas/handlers/create_document_reference_handler.py
```diff
@@ -152,6 +156,12 @@ def lambda_handler(event, context):
             500, "An error occurred when creating document reference", "POST"
         ).create_api_gateway_response()
         return response
+    except (LGInvalidFilesException, LGFileTypeException, LGFileNameException) as e:
+        logger.error(e)
+        response = ApiGatewayResponse(
+            400, "One or more if the files is not valid", "POST"
```
AlexHerbertNHS, line 162: I feel this should be more descriptive to give the user the info they need to fix the problem and try again. Since we're catching several different types of exception here, can we instead have a different error message depending on which exception is raised? e.g. 

File names do not fit into the pattern XofX_Lloyd_George_Record_[Patient Name]_[nhs number]_[DD-MM-YYYY].pdf
File type is incorrect- please ensure all Lloyd George files are PDF

and so on

lambdas/services/lloyd_george_validator.py
```diff
@@ -1,7 +1,44 @@
+import re
+
+from lambdas.models.nhs_document_reference import NHSDocumentReference
+
+
 class LGFileTypeException(ValueError):
-    """One or more of the files do not match the required file type."""
+    """One or more of the files do not match the required file type"""
```
AlexHerbertNHS, line 7: What's the """ do differently to "?

lambdas/services/lloyd_george_validator.py
```diff
@@ -1,7 +1,44 @@
...
         raise LGFileTypeException
+
+def validate_file_name(name: str):
+    lg_regex = r'[0-9]+of[0-9]+_Lloyd_George_Record_\[[A-Za-z ]+]_\[[0-9]{10}]_\[\d\d-\d\d-\d\d\d\d].pdf'
+    if not re.fullmatch(lg_regex, name):
+        raise LGFileNameException
+
+def check_for_duplicate_files(file_list: list[str]):
+    if len(file_list) > len(set(file_list)):
+        raise LGInvalidFilesException
```
AlexHerbertNHS, line 29: Instead of 3 seperate exceptions for this, would it not be better to make a single LGValidationException and allow it to be raised with a message inside detailing the specific failure which could be used to inform the user as noted above?

AlexHerbertNHS (changes requested):

lambdas/services/lloyd_george_validator.py
AlexHerbertNHS: Will this A-Za-z include special characters such as e, è, é, ê, ë for people with names that include these?

lambdas/services/lloyd_george_validator.py
```diff
@@ -1,7 +1,44 @@
...
+class LGInvalidFilesException(Exception):
+    """One or more of the files are not valid"""
+    pass
+
+def validate_lg_file_type(file_type: str):
     if file_type != 'application/pdf':
         raise LGFileTypeException
+
+def validate_file_name(name: str):
+    lg_regex = r'[0-9]+of[0-9]+_Lloyd_George_Record_\[[A-Za-z ]+]_\[[0-9]{10}]_\[\d\d-\d\d-\d\d\d\d].pdf'
```
AlexHerbertNHS, line 23: Will this A-Za-z include special characters such as e, è, é, ê, ë for people with names that include these?

lambdas/services/lloyd_george_validator.py
NogaNHS: I've changed it to one

lambdas/services/lloyd_george_validator.py
NogaNHS: I'm not sure if we want to include those. I'll check

lambdas/handlers/create_document_reference_handler.py
AlexHerbertNHS: That's fine if it's done in the UI, as long as the user isn't going to get a message saying "400, "One or more if the files is not valid" without telling them what they need to change to make it valid - if the file names are invalid I feel we should tell them why so they can fix it and try again

Additionally, this should be 'of', not 'if' - just seeing that now!

lambdas/services/lloyd_george_validator.py
AlexHerbertNHS: How about people with hyphenated names like Catherine Zeta-Jones or Sophie Ellis-Bextor, this will not match those:

1of2_Lloyd_George_Record_[Sophie Ellis-Bextor]_[1234567890]_[25-12-2019].pdf
1of2_Lloyd_George_Record_[Harvey Guillén]_[1234567890]_[25-12-2019].pdf

This SO looks promising:

https://stackoverflow.com/questions/24676691/whats-a-good-regex-to-include-accented-characters-in-a-simple-way

so I did one myself, this one matches both edge cases:

[0-9]+of[0-9]+_Lloyd_George_Record_\[[A-Za-z À-ÿ\-]+]_\[[0-9]{10}]_\[\d\d-\d\d-\d\d\d\d]

AlexHerbertNHS: Would be nice to see tests for those cases too :)

AlexHerbertNHS: `[0-9]+of[0-9]+_Lloyd_George_Record_\[[A-Za-z À-ÿ\-]+]_\[[0-9]{10}]_\[\d\d-\d\d-\d\d\d\d]`

^ since the above example was autoformatted by github to _italicise_ things in between 2 _'s

joefong-nhs: After the refinement session today, I suggest we do something like this to extract all wanted info first:

```
def extract_info_from_filename(filename: str) -> dict:
    page_number = r"(?P<page_no>[1-9][0-9]*)"
    total_page_number = r"(?P<total_page_no>[1-9][0-9]*)"
    patient_name = r"(?P<patient_name>[A-Za-z À-ÿ\-]+)"
    nhs_number = r"(?P<nhs_number>\d{10})"
    date_of_birth = r"(?P<dob>\d\d-\d\d-\d\d\d\d)"
    
    lg_regex = rf"{page_number}of{total_page_number}_Lloyd_George_Record_\[{patient_name}\]_\[{nhs_number}\]_\[{date_of_birth}\].pdf"

    if match := re.fullmatch(lg_regex, filename):
        return match.groupdict()
    else:
        raise LGInvalidFilesException('One or more of the files do not match naming convention')
```

Motive of having this:
1. Relatively easy to change the regex of one part in the filename e.g. patient name, date of birth
2. The logic of checking page numbers can be separately from the logic of regex matching
3. Relatively easy to add more further validation e.g. check every file got the same patient name, check date of birth is a valid date etc

NogaNHS: Done

AlexHerbertNHS (approved):
Looking good, thank you!

PR #87, opened 2023-10-11
PRMDR-264 View a Lloyd George Document e2e tests

carlsmith101:


app/cypress/e2e/0-ndr-core-tests/gp_view_lloyd_george_workflow.cy.js
```diff
@@ -0,0 +1,115 @@
...
+
+const baseUrl = Cypress.env('CYPRESS_BASE_URL') ?? 'http://localhost:3000/';
+
+describe('GP View Lloyd George Workflow', () => {
+    beforeEach(() => {
+        // Arrange
+        navigateToLgPage();
+    });
+
+    it.only('allows a GP user to view the Lloyd George document of an active patient', () => {
```
premnhs4, line 12: any specific reason, only this test is executed. Should we mention as smoke test (since we still use mock data)

Makefile
```diff
@@ -93,4 +93,4 @@ docker-down:
 	docker-compose -f ./app/docker-compose.yml down
 
 cypress-open:
-	npm --prefix ./app run cypress
+	TZ=GMT npm --prefix ./app run cypress
```
premnhs4, line 96: Will this work with the BST time change ?

app/cypress/support/e2e.ts
```diff
@@ -14,7 +15,19 @@
 // ***********************************************************
 
 // Import commands.js using ES2015 syntax:
-import './commands'
+import './commands';
+
+Cypress.Commands.add('getCy', (selector, ...args) => {
```
premnhs4, line 20: really helpful..LGTM

premnhs4 (commented):
LGTM : 
- Cleaned unwanted Cypress folder
- Introduced custom commands.
- just couple of queries left.
- Not sure why files in src was updated ( ie, pdfViewer.tsx & LloydGeorgeRecordPage.tsx) ?

app/cypress/e2e/0-ndr-core-tests/gp_view_lloyd_george_workflow.cy.js
carlsmith101: my mistake - I was meant to take that out. I included it for ease of creating the test. Will amend!

Makefile
carlsmith101: Annoyingly, I don't believe there is a way to enforce the timezone within Cypress config other than setting the 'TZ' environment variable.

I think we would have to add this as an environment variable into the github action pipeline jobs in order to ensure that the timezone is always aligned between local and pipeline runs.

Failing that, I could just amend the test assertion slightly so that it doesn't assert on the 'lastUpdated' time.

Any thoughts/suggestions?

premnhs4: Yeah ok..yes we need to look into the options of the env variable in github actions or even part of scripts section in package.json. we can live this for now.

PR #95, opened 2023-10-17
Prmdr 327 subtask -connect to PDS API

NogaNHS:


lambdas/handlers/search_patient_details_handler.py
```diff
@@ -1,15 +1,22 @@
...
+
+from services.mock_pds_service import MockPdsApiService
+
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
-
+def get_pds_service():
+    return PdsApiService if not bool(os.getenv("PDS_FHIR_IS_STUBBED")) else MockPdsApiService
 
 def lambda_handler(event, context):
```
abbas-khan10, line 21: Maybe worth updating this to match the other lambdas by using the decorators for validation, then we can remove the need to validate nhs number in the mock pds service

abbas-khan10, line 21: Maybe worth updating this to match the other lambdas by using the decorators for validation, then we can remove the need to validate nhs number in the mock pds service

lambdas/services/pds_api_service.py
```diff
@@ -1,23 +1,43 @@
...
+    def __init__(self, ssm_service):
+        self.ssm_service = ssm_service
 
-        return self.handle_response(response, nhs_number)
+    def fetch_patient_details(
+        self,
+        nhs_number: str,
+    ) -> PatientDetails:
+        try:
+            validate_id(nhs_number)
```
abbas-khan10, line 31: This can be removed too if validation is done in the lambda handler

lambdas/services/pds_api_service.py
```diff
@@ -1,23 +1,43 @@
...
+            response = self.pds_request(nhs_number, retry_on_expired=True)
+            return self.handle_response(response, nhs_number)
+        except ClientError as e:
+            logger.error(f"Error when getting ssm parameters {e}")
+            raise PdsErrorException("Failed to preform patient search")
 
     def handle_response(self, response: Response, nhs_number: str) -> PatientDetails:
         if response.status_code == 200:
-            patient = Patient.model_validate(response.content)
+            patient = Patient.model_validate(response.json())
```
abbas-khan10, line 40: Might have missed the implementation when I wrote the mock service but is there a check for if the `model_validate `fails? Think it raises a `ValidationError`

lambdas/services/pds_api_service.py
```diff
@@ -31,32 +51,100 @@ def handle_response(self, response: Response, nhs_number: str) -> PatientDetails
...
-            with open("services/mock_data/pds_patient.json") as f:
-                mock_pds_results.append(json.load(f))
-
-            with open("services/mock_data/pds_patient_restricted.json") as f:
-                mock_pds_results.append(json.load(f))
-
-        except FileNotFoundError:
-            raise PdsErrorException("Error when requesting patient from PDS")
+    def pds_request(self, nshNumber: str, retry_on_expired: bool):
+        endpoint, access_token_response = self.get_parameters_for_pds_api_request()
```
abbas-khan10, line 55: Minor gripe, feel free to ignore but what would you say to creating a pydantic model to represent the access_token_response? This way we can validate the response matches the model and don't need to stress about whether the dict keys we access below are present and try/catching `KeyError` exceptions

lambdas/services/pds_api_service.py
```diff
@@ -31,32 +51,100 @@ def handle_response(self, response: Response, nhs_number: str) -> PatientDetails
...
+        parameter_key = "/prs/dev-ndr/pds-fhir-access-token"
+        self.ssm_service.update_ssm_parameter(
+            parameter_key=parameter_key,
+            parameter_value=parameter_value,
+            parameter_type="SecureString",
+        )
+
+    def get_parameters_for_pds_api_request(self):
+        parameters = [
+            "/prs/dev/user-input/pds-fhir-endpoint",
```
abbas-khan10, line 117: Lots of string paths in this file that might be better stored away as constants in a config file somewhere to avoid hard coding them?

lambdas/services/pds_api_service.py
NogaNHS: The lambda will catch ValidationError

lambdas/services/pds_api_service.py
NogaNHS: I've created an enum class with all of the paths, is that what you had in mind?

abbas-khan10: Yep perfect

abbas-khan10 (approved):

lambdas/handlers/search_patient_details_handler.py
AlexHerbertNHS: 👍 to this - just add the following above the method declaration:

`@validate_patient_id`

lambdas/services/mock_pds_service.py
```diff
@@ -0,0 +1,66 @@
...
+            patient_details = patient.get_patient_details(nhs_number)
+            return patient_details
+
+        if response.status_code == 404:
+            raise PatientNotFoundException(
+                "Patient does not exist for given NHS number"
+            )
+
+        if response.status_code == 400:
+            raise InvalidResourceIdException("Invalid NHS number")
```
AlexHerbertNHS, line 34: This should be caught and handled while validating the NHS number at the start of the handler. Not saying that we won't get a 400 error back from PDS, but perhaps it would be best to return the PDS response body in the Exception body, as opposed to just assuming that it's because the NHS number is invalid, I think

lambdas/handlers/search_patient_details_handler.py
```diff
@@ -29,10 +45,7 @@ def lambda_handler(event, context):
     except PatientNotFoundException as e:
         return ApiGatewayResponse(404, f"{str(e)}", "GET").create_api_gateway_response()
 
-    except InvalidResourceIdException as e:
-        return ApiGatewayResponse(400, f"{str(e)}", "GET").create_api_gateway_response()
-
-    except PdsErrorException as e:
+    except (InvalidResourceIdException, PdsErrorException) as e:
```
AlexHerbertNHS, line 48: Should we log these to CloudWatch?

lambdas/services/mock_pds_service.py
AlexHerbertNHS: Ah, this is for the fake service. Not sure if what I've written is still valid - sorry!

lambdas/handlers/search_patient_details_handler.py
NogaNHS: Done

PR #101, opened 2023-10-19
PRMDR: Removed Old Lambda Actions and amended titles to sort nicely in the actions tab

SRAlexander:


.github/workflows/new_base-lambdas-reusable-deploy-all.yml
```diff
@@ -1,4 +1,4 @@
-name: "BASE: Deploy all required Lambda functions"
+name: "ZBASE Lambdas: Deploy all required Lambda functions"
```
abbas-khan10, line 1: What is the purpose of `Z-BASE`?

abbas-khan10:
If the old files are being deleted, is there any reason why we're still keeping the `new_` prefix on the replacement filenames?

abbas-khan10 (approved):

SRAlexander:
The new prefix is remaining for the moment as there are a lot of references across both the infrastructure and this repo. It can be done but the priority of this ticket is to remove the duplication of the actions.

PR #103, opened 2023-10-19
Prmdr 168 Back Channel Logout

thisusernameisnowtaken:


lambdas/handlers/back_channel_logout_handler.py
```diff
@@ -0,0 +1,62 @@
...
+        logger.info("decoding token")
+        oidc_service = OidcService()
+        decoded_token = oidc_service.validate_and_decode_token(token)
+        session_id = decoded_token["sid"]
+        remove_session_from_dynamo_db(session_id)
+
+    except ClientError as e:
+        logger.error(f"Error logging out user: {e}")
+        return ApiGatewayResponse(
+            400, """{ "error":"Internal error logging user out"}""", "POST"
```
NogaNHS, line 40: Shouldn't we return 500 here?

lambdas/handlers/back_channel_logout_handler.py
```diff
@@ -0,0 +1,62 @@
...
+        return ApiGatewayResponse(
+            400, """{ "error":"No sid field in decoded token"}""", "POST"
+        ).create_api_gateway_response()
+
+    return ApiGatewayResponse(200, "", "POST").create_api_gateway_response()
+
+
+def remove_session_from_dynamo_db(session_id):
+    logger.info(f"Session to be removed: {session_id}")
+    dynamodb_name = os.environ["AUTH_DYNAMODB_NAME"]
```
NogaNHS, line 58: This env var is missing from '@ensure_environment_variables'

SRAlexander (approved):

lambdas/handlers/back_channel_logout_handler.py
AlexHerbertNHS: Done

lambdas/handlers/back_channel_logout_handler.py
AlexHerbertNHS: Done

PR #104, opened 2023-10-19
Prmdr sonarcube coverage

SRAlexander:


carlsmith101 (changes requested):
the python code coverage currently includes the python test files too. We can omit these by adding the tests dir to a .coveragerc file:
```
# lambdas/.coveragerc
[run]
omit = tests/*
```

carlsmith101 (approved):

PR #108, opened 2023-10-23
PRMDR-311: Delete doc reference lambda and document service refactor

abbas-khan10:


lambdas/enums/metadata_field_names.py
```diff
@@ -2,32 +2,30 @@
...
+            DocumentReferenceMetadataFields.ID.value,
+            DocumentReferenceMetadataFields.CONTENT_TYPE.value,
+            DocumentReferenceMetadataFields.CREATED.value,
+            DocumentReferenceMetadataFields.DELETED.value,
+            DocumentReferenceMetadataFields.FILE_NAME.value,
+            DocumentReferenceMetadataFields.FILE_LOCATION.value,
+            DocumentReferenceMetadataFields.NHS_NUMBER.value,
+            DocumentReferenceMetadataFields.TTL.value,
+            DocumentReferenceMetadataFields.TYPE.value,
+            DocumentReferenceMetadataFields.VIRUS_SCANNER_RESULT.value,
```
joefong-nhs, line 28: If we are not skipping some field, I think we can get all the field value like this:
```
        return [field.value for field in DocumentReferenceMetadataFields]
```

lambdas/models/document_reference.py
```diff
@@ -0,0 +1,86 @@
...
+                return file_key
+            raise InvalidDocumentReferenceException(
+                "Failed to parse object key from file location"
+            )
+        except IndexError:
+            raise InvalidDocumentReferenceException(
+                "Failed to parse object key from file location"
+            )
+
+    def create_unique_filename(self, duplicates):
```
joefong-nhs, line 70:
```suggestion
    def create_unique_filename(self, duplicates: int):
```
From the usage seems that this `duplicates` is an integer

joefong-nhs (approved):
Looks great to me! Really appreciate your hard work in migrating the model to pydantic and all the refactoring 😄  
Gave a few minor suggestions, but I think it's good to go without any change :)

lambdas/services/document_manifest_service.py
```diff
@@ -5,7 +5,7 @@
 import zipfile
 
 from botocore.exceptions import ClientError
-from models.document import Document
+from models.document_reference import DocumentReference
```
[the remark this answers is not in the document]
abbas-khan10: Good spot, forgot to delete this after migrating the functionality

PR #109, opened 2023-10-24
Prdmr 370 - LG validation using pds service

NogaNHS:


lambdas/tests/unit/helpers/data/pds/pds_patient_response.py
```diff
@@ -63,7 +63,7 @@
             "identifier": {
                 "system": "https://fhir.nhs.uk/Id/ods-organization-code",
                 "value": "Y12345",
-                "period": {"start": "2020-01-01", "end": "2021-12-31"},
+                "period": {"start": "2020-01-01"},
```
SRAlexander, line 66: Are there no cases with an end date include now?

lambdas/utils/lloyd_george_validator.py
```diff
@@ -87,3 +96,38 @@ def check_for_file_names_agrees_with_each_other(file_name_list: list[str]):
...
+        current_user_ods = get_user_ods_code()
+        if patient_details.general_practice_ods != current_user_ods:
+            raise LGInvalidFilesException("User is not allowed to access patient")
+
+    except (HTTPError, ValidationError, ClientError, ValueError) as e:
+        logger.error(e)
+        raise LGInvalidFilesException(e)
+
+def get_user_ods_code():
+    ssm_service = SSMService()
```
SRAlexander, line 132: The fact we're calling into a custom service means this is more fitting as a service than a util class. It does not look like you've implemented this class as part of this ticket so there's no action but this is one we should raise in our clean code activity. :)

SRAlexander (approved):

lambdas/utils/lloyd_george_validator.py
```diff
@@ -87,3 +96,38 @@ def check_for_file_names_agrees_with_each_other(file_name_list: list[str]):
     ]
     if len(set(expected_common_part)) != 1:
         raise LGInvalidFilesException("File names does not match with each other")
+
+def validate_with_pds_service(file_name_list: list[str], nhs_number: str):
+    try:
+    # Check file names match with the nhs number in metadata.csv
```
joefong-nhs, line 102: this comment is outdated so probably better to change it or delete it.

lambdas/utils/lloyd_george_validator.py
```diff
@@ -87,3 +96,38 @@ def check_for_file_names_agrees_with_each_other(file_name_list: list[str]):
...
+        patient_full_name =' '.join([name for name in patient_details.given_Name]) + ' ' + patient_details.family_name
+        if patient_full_name != patient_name:
+            raise LGInvalidFilesException("Patient name does not match our records")
+        current_user_ods = get_user_ods_code()
+        if patient_details.general_practice_ods != current_user_ods:
+            raise LGInvalidFilesException("User is not allowed to access patient")
+
+    except (HTTPError, ValidationError, ClientError, ValueError) as e:
+        logger.error(e)
+        raise LGInvalidFilesException(e)
```
joefong-nhs, line 129: Ideally I would suggest to put HTTPError and ClientError in another except block and handle them differently, because those are actually our internal problem rather than LG files being invalid, but we can leave that to future refactoring

lambdas/utils/lloyd_george_validator.py
```diff
@@ -87,3 +96,38 @@ def check_for_file_names_agrees_with_each_other(file_name_list: list[str]):
...
+
+        date_of_birth = datetime.datetime.strptime(date_of_birth, '%d-%m-%Y').date()
+        if patient_details.birth_date != date_of_birth:
+            raise LGInvalidFilesException("Patient DoB does not match our records")
+        patient_full_name =' '.join([name for name in patient_details.given_Name]) + ' ' + patient_details.family_name
+        if patient_full_name != patient_name:
+            raise LGInvalidFilesException("Patient name does not match our records")
+        current_user_ods = get_user_ods_code()
+        if patient_details.general_practice_ods != current_user_ods:
+            raise LGInvalidFilesException("User is not allowed to access patient")
```
joefong-nhs, line 125: Later maybe it's better to make the `current_user_ods` to an argument that got passed into this method, so that it is easier to extend to other GP. Anyway this is not urgent and can be skip for now.

lambdas/utils/lloyd_george_validator.py
NogaNHS: I can catch it in the lambda, instead of raising `LGInvalidFilesException`. How do you suggest handling them but logging the error and stopping validation?

joefong-nhs (approved):
Gave some minor suggestions, happy to approve this PR.

lambdas/utils/lloyd_george_validator.py
joefong-nhs: Agree. I think we can leave that for future refactoring.

lambdas/utils/lloyd_george_validator.py
NogaNHS: Done

lambdas/tests/unit/utils/test_lloyd_george_validator.py
```diff
@@ -117,3 +122,160 @@ def test_files_for_different_patients():
...
+        mock_pds_call = mocker.patch(
+            "services.pds_api_service.PdsApiService.pds_request"
+        )
+        mock_odc_code = mocker.patch("utils.lloyd_george_validator.get_user_ods_code", return_value='Y12345')
+
+        validate_with_pds_service(lg_file_list, "9000000009")
+
+        mock_pds_call.assert_not_called()
+        mock_odc_code.assert_not_called()
+
```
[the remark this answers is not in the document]
NogaNHS: Done

lambdas/models/pds_models.py
```diff
@@ -79,6 +91,13 @@ def get_current_home_address(self) -> Optional[Address]:
                 if entry.use.lower() == "home":
                     return entry
 
+    def get_ods_code_for_gp(self) -> str:
```
abbas-khan10, line 94: I know it's cheeky to ask as I didn't write any when I first made this class, but maybe some unit tests for the pds models now the number of functions has increased just to verify everything works as expected?

NogaNHS: I've added a few tests

abbas-khan10 (approved):

SRAlexander (approved):

PR #111, opened 2023-10-24
PRMDR-322 download Lloyd George cypress tests

carlsmith101:


app/src/components/layout/navLinks/NavLinks.tsx
```diff
@@ -22,7 +22,7 @@ const NavLinks = () => {
             <Header.NavItem
                 role="link"
                 className="clickable"
-                data-cy="logout-btn"
+                data-testid="logout-btn"
```
SRAlexander, line 25: All good as it is but don't feel the need to add test to the attribute, if it's an ID for a component go for it. :)

SRAlexander: Ahh I can see what you've done above, but Id should be enough.

app/cypress/e2e/0-ndr-core-tests/gp_view_lloyd_george_workflow.cy.js
```diff
@@ -17,77 +16,162 @@ describe('GP View Lloyd George Workflow', () => {
...
+                body: 'No documents found for given NHS number and document type',
+            }).as('documentManifest');
+
+            cy.getByTestId('actions-menu').click();
+            cy.getByTestId('download-all-files-link').click();
+
+            cy.wait('@documentManifest');
+
+            // Assert
+            // TODO - implement error scenario in UI and amend assertion accordingly
```
SRAlexander, line 140: Looks like a required case has not been implemented in dev. From a code perspective this can be merged but do we need to raise a bug to cover this?

SRAlexander (approved):

app/src/components/layout/navLinks/NavLinks.tsx
carlsmith101: I generally tend to prefer using `data-*` custom attributes where possible, as this is considered a best practice in cypress and other UI test frameworks - https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

app/cypress/e2e/0-ndr-core-tests/gp_view_lloyd_george_workflow.cy.js
carlsmith101: Sure, I've set these tests to skip for now, and will raise bug tickets for the fix

app/src/components/layout/navLinks/NavLinks.tsx
RioKnightleyNHS: Personally I think data-testid is the ideal attribute if we are going this route, as jest also has built in query selectors that use data-testid

app/src/components/blocks/lloydGeorgeRecordDetails/LloydGeorgeRecordDetails.tsx
```diff
@@ -15,6 +15,7 @@ export type Props = {
 
 type PdfActionLink = {
     label: string;
+    dataTestId: string;
```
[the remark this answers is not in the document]
RioKnightleyNHS: This approach is fine, but maybe if we can change `dataTestId` to a name that can be reusable as these are technically unique identifies. 

Maybe change the prop to be called `key` or `id` instead. Then we can share this prop on the key and data-testid part of the mapped link render instead of the concatenated (and ugly) key currently on the links

RioKnightleyNHS (changes requested):
See comments on dataTestId prop

app/src/components/blocks/lloydGeorgeRecordDetails/LloydGeorgeRecordDetails.tsx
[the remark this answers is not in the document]
carlsmith101: amended

RioKnightleyNHS (approved):

PR #112, opened 2023-10-25
[FIX] PRMDR-311: Delete docs lambda fix for optional TTL validation and return type

abbas-khan10:


lambdas/handlers/delete_document_reference_handler.py
```diff
@@ -47,13 +48,15 @@ def lambda_handler(event, context):
...
     except ClientError as e:
         logger.info(str(e))
         return ApiGatewayResponse(
             500, "Failed to delete documents", "DELETE"
         ).create_api_gateway_response()
 
-    return ApiGatewayResponse(
-        204, "Successfully deleted documents", "DELETE"
-    ).create_api_gateway_response()
+    return ApiGatewayResponse(200, "Success", "DELETE").create_api_gateway_response()
```
joefong-nhs, line 62: do we need to amend front end to check for 200 rather than 204?  
(ignore me if it is already done or planned in other ticket)

lambdas/services/document_service.py
```diff
@@ -89,12 +104,12 @@ def delete_documents(self, table_name: str, documents: list[DocumentReference]):
...
-                tag_key=str(S3LifecycleTags.SOFT_DELETE_KEY.value),
-                tag_value=str(S3LifecycleTags.SOFT_DELETE_VAL.value),
+                file_key=reference.get_file_key(),
+                s3_bucket_name=reference.get_file_bucket(),
+                tag_key=str(S3LifecycleTags.SOFT_DELETE.value),
+                tag_value=str(S3LifecycleTags.ENABLE_TAG.value),
             )
 
-            self.update_item(table_name, doc.id, updated_fields=update_fields)
+            self.update_item(table_name, reference.id, updated_fields=update_fields)
```
joefong-nhs, line 115: It appears to me that this part of code will put a soft_delete tag, even if type_of_delete is death_delete. Is that something we could have a look ?

joefong-nhs (commented):
gave some comment

lambdas/handlers/delete_document_reference_handler.py
abbas-khan10: We changed it so it could be consumed correctly in the FE, Rachel is aware and has made this change in the FE :)

lambdas/services/document_service.py
abbas-khan10: Good spot will change that now

joefong-nhs (approved):

PR #114, opened 2023-10-26
PRMDR-332 Delete all documents PCSE journey

RachelHowellNHS:


app/src/components/blocks/deleteDocumentsStage/DeleteDocumentsStage.tsx
```diff
@@ -65,16 +70,35 @@ function DeleteDocumentsStage({
...
+                        docType: DOCUMENT_TYPE.ARF,
+                        nhsNumber: patientNhsNumber,
+                        baseUrl,
+                        baseHeaders,
+                    });
+                    response = await deleteAllDocuments({
+                        docType: DOCUMENT_TYPE.LLOYD_GEORGE,
+                        nhsNumber: patientNhsNumber,
+                        baseUrl,
+                        baseHeaders,
```
RioKnightleyNHS, line 94: Not entirely sure if this is the implementation we are looking for if you are trying to check the status on multiple requests. There is a more complex but cleaner way of doing this by making an array of promises and reducing their outputs into one status code.

I was also looking for why we are assigning response when we only care about the status. At first it seemed like we were trying to fetch something more than just the status code. We can catchup in around about this block

app/src/components/blocks/deleteDocumentsStage/DeleteDocumentsStage.tsx
```diff
@@ -83,8 +107,17 @@ function DeleteDocumentsStage({
...
+            if (userType === USER_ROLE.GP) {
+                if (setStage) {
+                    setStage(LG_RECORD_STAGE.RECORD);
+                }
+            }
+            if (userType === USER_ROLE.PCSE) {
+                if (setIsDeletingDocuments) {
+                    setIsDeletingDocuments(false);
+                }
+            }
```
RioKnightleyNHS, line 120: There are quite alot of if else statements nested, including the big one for 'no' and 'yes' options, these can be seperated into more functions such as handleYesOption() handleNoOption()

Also maybe these no and yes option values can be enums? It's always better for type safety

RioKnightleyNHS (changes requested):

RioKnightleyNHS (approved):

PR #121, opened 2023-10-31
Prmdr 369 and 372

joefong-nhs:


lambdas/services/bulk_upload_service.py
```diff
@@ -29,8 +34,9 @@ def __init__(
         self.staging_bucket_name = os.environ["STAGING_STORE_BUCKET_NAME"]
         self.lg_bucket_name = os.environ["LLOYD_GEORGE_BUCKET_NAME"]
         self.lg_dynamo_table = os.environ["LLOYD_GEORGE_DYNAMODB_NAME"]
+        self.bulk_upload_report_dynamo_table = os.environ["BULK_UPLOAD_DYNAMODB_NAME"]
         self.invalid_queue_url = os.environ["INVALID_SQS_QUEUE_URL"]
-
+        self.metadata_queue_url = os.environ["METADATA_SQS_QUEUE_URL"]
```
NogaNHS, line 39: Are we catching KeyError somewhere (that is not in the general Exception in the lambda)?

lambdas/handlers/bulk_upload_handler.py
```diff
@@ -23,28 +18,7 @@ def lambda_handler(event, _context):
...
-    except KeyError:
-        nhs_number = ""
-
-    sqs_service.send_message_with_nhs_number_attr(
-        queue_url=invalid_queue_url,
-        message_body=json.dumps(new_message),
-        nhs_number=nhs_number,
-    )
-    logger.info(f"Sent message to invalid queue: {invalid_message}")
+        except Exception as error:
```
abbas-khan10, line 21: Should we be catching generic errors here? Think it's best practice to only catch what we know it could throw

joefong-nhs: Thanks!  Agree that it is bad to catch the generic error type.
I have changed this part to catch a list of several possible errors.

lambdas/models/config.py
```diff
@@ -7,3 +7,7 @@ def to_camel(string: str) -> str:
 
 
 conf = ConfigDict(alias_generator=to_camel)
+
+
+def to_capwords(snake_case_string: str) -> str:
+    return "".join(x.capitalize() for x in snake_case_string.split("_"))
```
abbas-khan10, line 13: I installed a package called inflection for a previous ticket which may come in use here rather than creating our own transformation functions on strings [](https://inflection.readthedocs.io/en/latest/) can use the camelize function from the library and then capitalize? Up to you though just thought you might be interested in taking a look at what it can do :)

joefong-nhs: Thanks!  It worked nicely as a drop in replacement :D   A great util library to know
I have updated this part to use inflection

lambdas/services/bulk_upload_service.py
joefong-nhs: No, because this lambda isn't a web api, so I don't put anything to catch missing env var here.
If any env var is missing from here, the lambda would fail immediately during the `bulk_upload_service = BulkUploadService()`

lambdas/models/config.py
```diff
@@ -7,3 +7,7 @@ def to_camel(string: str) -> str:
 
 
 conf = ConfigDict(alias_generator=to_camel)
+
+
+def to_capwords(snake_case_string: str) -> str:
+    return inflection.camelize(snake_case_string, uppercase_first_letter=True)
```
abbas-khan10, line 13: Quick question on this function, seems it's doing the same as the `to_camel` function above. What's the output we are looking for here that makes it different?

joefong-nhs: `to_camel` has uppercase_first_letter=False and this one got uppercase_first_letter=True . 
I tried `to_camel` but it doesn't fit the current schema we are using so made this one.

abbas-khan10: Ah I see you're right, in that case maybe change the name of the function to `to_capitalized_camel`? `capwords` suggests all letters are capital that's all but I'll leave that up to you

abbas-khan10 (approved):

lambdas/models/config.py
joefong-nhs: Nice one. Agree that `capwords` is kind of confusing. Renamed to `to_capitalized_camel` :D

PR #124, opened 2023-11-01
[FIX]PRMDR-427: Refactor delete lambda to handle multiple document types in query params

abbas-khan10:


lambdas/handlers/delete_document_reference_handler.py
```diff
@@ -25,38 +25,70 @@
...
-        if SupportedDocumentTypes.LG.name == doc_type:
-            logger.info("Retrieving Lloyd George documents for deletion")
-            table = os.environ["LLOYD_GEORGE_DYNAMODB_NAME"]
+            if (
+                arf_delete_response["statusCode"] == 404
+                and lg_delete_response["statusCode"] == 404
+            ):
+                return ApiGatewayResponse(
+                    404, "No documents available", "DELETE"
+                ).create_api_gateway_response()
```
joefong-nhs, line 47: just a little suggestion, would it be good to change `extract_document_type` to return the enum type instead of enum.value string ? 
so that we could use directly use doc_type as our custom enum type within our boundary, and don't need to do a .value every time when we compare
Although I've read your tests cases and I think that current implementation should be working fine :) 

Or, maybe we can put this as a little side note for future refactoring 
(possibly after we bump up github action to python 3.11, when we can use StrEnum)

lambdas/tests/unit/services/test_document_service.py
```diff
@@ -30,59 +31,74 @@ def nhs_number():
...
+    ]
+
+    result = service.fetch_available_document_references_by_type(TEST_NHS_NUMBER, "ALL")
+
+    assert mock_dynamo_table.call_count == 2
+
+    assert len(result) == 6
+    assert type(result[0]) == DocumentReference
+
+
```
joefong-nhs, line 87: suggest that we can add a unit test when we fetching ALL but one type have record and the other type is empty response:
 
```suggestion

def test_returns_list_of_both_type_document_references_when_only_one_type_have_result(
    set_env, mocker
):
    mocker.patch("boto3.resource")
    service = DocumentService()

    mock_dynamo_table = mocker.patch.object(service.dynamodb, "Table")
    mock_dynamo_table.return_value.query.side_effect = [
        MOCK_SEARCH_RESPONSE,
        MOCK_EMPTY_RESPONSE,
    ]

    result = service.fetch_available_document_references_by_type(TEST_NHS_NUMBER, "ALL")

    assert mock_dynamo_table.call_count == 2

    assert len(result) == 3
    assert type(result[0]) is DocumentReference

```

lambdas/tests/unit/services/test_document_service.py
```diff
@@ -30,59 +31,74 @@ def nhs_number():
...
 
-        mock_dynamo.return_value.Table.assert_called_with(expected_table)
+    result = service.fetch_available_document_references_by_type(TEST_NHS_NUMBER, "LG")
 
-        assert len(result) == 3
-        assert type(result[0]) == DocumentReference
+    mock_dynamo_table.assert_called_with(MOCK_LG_TABLE_NAME)
 
+    assert len(result) == 3
+    assert type(result[0]) == DocumentReference
```
joefong-nhs, line 48: for this line, my linter suggested to change to `assert type(result[0]) is DocumentReference` or `assert isinstance(result[0], DocumentReference)` 
Personally I think in our use case using == to compare isn't a problem so up to you :)

joefong-nhs (approved):
Looks great to me. The parametrize tests are particularly nicely done. 👍 
Gave some minor suggestions but I think this is good enough in current form :)

lambdas/handlers/delete_document_reference_handler.py
abbas-khan10: Yeah I agree think this may need revisiting later on down the line as you've mentioned StrEnums would be great to implement across the whole app not just for SupportedDocTypes. I've added this to my refactoring notes

lambdas/enums/supported_document_types.py
```diff
@@ -2,24 +2,20 @@
...
+    ALL = "ALL"
 
     @staticmethod
     def list():
         return [SupportedDocumentTypes.ARF, SupportedDocumentTypes.LG]
 
     @staticmethod
-    def list_names():
-        return [SupportedDocumentTypes.ARF.name, SupportedDocumentTypes.LG.name]
+    def list_names() -> [str]:
```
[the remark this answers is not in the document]
abbas-khan10: Good spot that's a typo

lambdas/tests/unit/services/test_document_service.py
abbas-khan10: No I agree, cleaner test for equality

PR #158, opened 2023-11-22
PRMDR-469 Refactor authoriser handler

NogaNHS:
Co-authored-by:  @RachelHowellNHS

### DIFF
--- a/lambdas/handlers/authoriser_handler.py
+++ b/lambdas/handlers/authoriser_handler.py
@@ -54,8 +54,7 @@ def lambda_handler(event, context):
             policy.denyMethod(_http_verb, path)
 
     except AuthorisationException as e:
-        logger.error(e)
-        logger.error("failed to authenticate user")
+        logger.error(f"failed to authenticate user due to: {e}")
         policy.denyAllMethods()
 
     auth_response = policy.build()
--- a/lambdas/handlers/authoriser_handler.py
+++ b/lambdas/handlers/authoriser_handler.py
@@ -44,9 +44,7 @@ def lambda_handler(event, context):
     policy.stage = stage
     try:
         authoriser_service = AuthoriserService()
-        logger.info(
-            "Validating resource req: %s, http: %s" % (path, _http_verb)
-        )
+        logger.info("Validating resource req: %s, http: %s" % (path, _http_verb))
         is_allow_policy = authoriser_service.auth_request(
             path, ssm_jwt_public_key_parameter, auth_token
         )
--- a/lambdas/handlers/authoriser_handler.py
+++ b/lambdas/handlers/authoriser_handler.py
@@ -32,7 +32,7 @@ def lambda_handler(event, context):
     logger.info("Authoriser handler triggered")
     ssm_jwt_public_key_parameter = os.environ["SSM_PARAM_JWT_TOKEN_PUBLIC_KEY"]
     auth_token = event.get("authorizationToken")
-    principal_id = ""
+    principal_id = context.authorizer.principalId
     if event.get("methodArn") is None:
         return {"Error": "methodArn is not defined"}
     _, _, _, region, aws_account_id, api_gateway_arn = event.get("methodArn").split(":")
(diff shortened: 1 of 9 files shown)


lambdas/handlers/authoriser_handler.py
```diff
@@ -34,126 +28,35 @@
...
-            else:
-                policy.denyMethod(_http_verb, path)
-
-        auth_response = policy.build()
-        return auth_response
+            policy.denyMethod(_http_verb, path)
 
     except AuthorisationException as e:
         logger.error(e)
         logger.error("failed to authenticate user")
```
abbas-khan10, line 58: Is there a need for 2 logs here? Wouldn't the exception message be sufficient? Or maybe combine them as 1 log?

lambdas/models/auth_policy.py
```diff
@@ -36,8 +36,8 @@ class AuthPolicy(object):
     region = os.getenv("AWS_DEFAULT_REGION", "eu-west-2")
     stage = "dev"
 
-    def __init__(self, principal, awsAccountId):
-        self.awsAccountId = awsAccountId
+    def __init__(self, principal, aws_account_id):
+        self.awsAccountId = aws_account_id
         self.principalId = principal
```
abbas-khan10, line 41: I know this code existed before this PR, but there's camel case variables in this class that should probably be changed to snake case as part of the refactor

lambdas/services/authoriser_service.py
```diff
@@ -0,0 +1,106 @@
...
+from utils.utilities import redact_id_to_last_4_chars
+
+logger = LoggingService(__name__)
+
+
+class AuthoriserService:
+    def __init__(
+        self,
+        path,
+        _http_verb,
```
abbas-khan10, line 22: Any reason for the _ in this variable name?

lambdas/handlers/authoriser_handler.py
```diff
@@ -34,126 +28,35 @@
 @ensure_environment_variables(names=["SSM_PARAM_JWT_TOKEN_PUBLIC_KEY"])
 @override_error_check
 def lambda_handler(event, context):
+    request_context.app_interaction = LoggingAppInteraction.LOGIN.value
+    logger.info("Authoriser handler triggered")
+    ssm_jwt_public_key_parameter = os.environ["SSM_PARAM_JWT_TOKEN_PUBLIC_KEY"]
+    auth_token = event.get("authorizationToken")
+    principal_id = ""
```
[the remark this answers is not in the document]
NogaNHS: Not sure if this is a placeholder for a future work. But it was already there so we didn't change it.

lambdas/handlers/authoriser_handler.py
NogaNHS: Done

lambdas/handlers/authoriser_handler.py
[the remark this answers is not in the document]
abbas-khan10: Might be worth finding out from @joefong-nhs maybe, or whoever wrote this originally as it could have been something missed in previous PR's

lambdas/services/authoriser_service.py
NogaNHS: This only being used in a log really. and it was defined as _** is the original code.

lambdas/models/auth_policy.py
NogaNHS: Done

lambdas/handlers/authoriser_handler.py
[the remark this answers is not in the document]
joefong-nhs: We got the basis of our authoriser code from AWS Blueprint. In the blueprint code, this principal id was hard-coded as an arbitrary value  `principalId = "user|a1b2c3d4"`, so we just replaced that with an empty string.
I think in our use case this principal_id can be any string. We could replace this with a proper principal id if we got one.

abbas-khan10 (approved):

PR #159, opened 2023-11-23
PRMDR-523 fix log out link

RachelHowellNHS:
currently deployed to ndrc, can look at ndr-test to see the issue and ndrc for the fix

### DIFF
--- a/lambdas/handlers/logout_handler.py
+++ b/lambdas/handlers/logout_handler.py
@@ -2,7 +2,6 @@ import os
 
 import boto3
 import jwt
-import json
 from botocore.exceptions import ClientError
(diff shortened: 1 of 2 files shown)


lambdas/handlers/logout_handler.py
```diff
@@ -22,7 +21,8 @@ def lambda_handler(event, context):
     token = None
     headers = event.get("headers")
     if headers:
-        token = headers["x-auth"]
+        print(headers)
```
SRAlexander, line 24: Can we remove the log line here please?

RachelHowellNHS: done

PR #161, opened 2023-11-24
PRMDR-517 Catch 429 'Too Many Requests' error from PDS

joefong-nhs:
- Catch 429 'Too Many Requests' response from PDS at LG validator
- Add logic at BulkUploadLamda to put unprocessed messages back to sqs queue and retry after 5 minutes
- Add nhs number to logging context for BulkUploadLamda
- In the mock PDS service, add a switch to allow mocking 429 'Too Many Requests' error in sandbox (just for testing purpose. should be inactive unless manually set the env var to true)


thisusernameisnowtaken (approved):
All looks good. Some improvements in code structure can be made but would be time consuming and is covered by the refactoring tickets.

PR #162, opened 2023-11-24
PRMDR-518 FIFO queue change

NogaNHS:


lambdas/services/bulk_upload_service.py
```diff
@@ -192,6 +193,7 @@ def put_message_back_to_queue(self, staging_metadata: StagingMetadata):
             message_body=staging_metadata.model_dump_json(by_alias=True),
             nhs_number=staging_metadata.nhs_number,
             delay_seconds=60 * 5,
+            group_id= f"back_to_queue_bulk_upload_{uuid.uuid4()}"
```
abbas-khan10, line 196: Has there been a discussion about what these group ID's should be? Are they required anywhere or are we specifying them here as they don't have a default? (Same for the usages below)

NogaNHS: In a FIFO queue we're required to pass a group id, but it's for our use. It should be the same for all messages sent in a lambda run. Do you have concerns about it?

AlexHerbertNHS (approved):

lambdas/services/bulk_upload_service.py
abbas-khan10: No concerns just wondering what the decision for the ID was but if it's for our purpose that's fine by me :)

abbas-khan10 (approved):

PR #167, opened 2023-11-27
Prmdr cypress reports

SRAlexander:


RioKnightleyNHS (approved):

app/cypress/e2e/0-ndr-core-tests/auth_routes/auth_general_browser_states.cy.js
```diff
@@ -17,11 +17,11 @@ describe('authentication & authorisation', () => {
...
-
-            assertSessionStorage({
-                auth: null,
-                isLoggedIn: false,
+            cy.wait('@logout').then(() => {
+                assertSessionStorage({
+                    auth: null,
+                    isLoggedIn: false,
+                });
             });
```
RioKnightleyNHS, line 25: Added to fix Edge errors. Recommend that this is how old cy.wait() callbacks are handled to reduce this async problem

app/cypress/e2e/0-ndr-core-tests/auth_routes/auth_general_browser_states.cy.js
```diff
@@ -17,11 +17,11 @@ describe('authentication & authorisation', () => {
                 statusCode: 200,
             }).as('logout');
             cy.getByTestId('logout-btn').click();
-            cy.wait('@logout');
-
-            assertSessionStorage({
-                auth: null,
-                isLoggedIn: false,
+            cy.wait('@logout').then(() => {
```
RioKnightleyNHS, line 20: Cypres documentation says using .then after a cy.wait() is an anti-pattern. But for some reason Edge fails on this particular case without it

RioKnightleyNHS (approved):

PR #169, opened 2023-11-28
PRMDR 387 fix to process grave accent alphabets

NogaNHS:


lambdas/handlers/bulk_upload_handler.py
```diff
@@ -43,6 +42,7 @@ def lambda_handler(event, _context):
             TypeError,
             AttributeError,
         ) as error:
-            logger.info(f"Fail to process current message due to error: {error}")
+            logger.error("Fail to process current message due to error:")
```
SRAlexander, line 45: Should this not just be one error message?

SRAlexander (approved):

PR #171, opened 2023-11-28
PRMDR-473 refactor bulk upload report handler

RachelHowellNHS:


lambdas/services/bulk_upload_report_service.py
```diff
@@ -0,0 +1,118 @@
...
+                file_key=f"reports/{file_name}",
+                file_name=f"/tmp/{file_name}",
+            )
+        except ClientError as e:
+            logger.error(str(e))
+            return LambdaResponse(
+                status=400,
+                message="Bulk upload report creation failed",
+                errors=[LambdaResponseError(type="S3 Client", message=str(e))],
+            )
```
NogaNHS, line 63: How are we returning this 400 to? 😮

lambdas/handlers/bulk_upload_report_handler.py
```diff
@@ -25,82 +17,5 @@
...
-        end_timestamp -= datetime.timedelta(days=1)
-    start_timestamp = end_timestamp - datetime.timedelta(days=1)
-    return start_timestamp, end_timestamp
-
-
-def write_empty_report(file_path: str):
-    with open(file_path, "w") as output_file:
-        output_file.write("No data was found for this timeframe")
+    response = bulk_upload_report_service.report_handler()
+    return response.build_response()
```
joefong-nhs, line 21: As I remember, the bulk upload related lambdas are triggered by cron jobs or SQS, so probably this response would not be handled by any entity. 
Therefore, I think it might be better to raise an error directly if it fails, rather than returning with http status code (200, 400 etc)

NogaNHS: I was thinking the same but the EventBridge trigger an API endpoint for this

lambdas/utils/exceptions.py
```diff
@@ -82,3 +82,7 @@ class S3FileNotFoundException(Exception):
 
 class TagNotFoundException(Exception):
     pass
+
+
+class BulkUploadReportException(Exception):
+    pass
```
NogaNHS, line 88: Is this being used?

lambdas/services/bulk_upload_report_service.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return LambdaResponse(
+            status=200, message="Bulk upload report creation successful"
+        )
```
NogaNHS, line 67: Can the lambda response be moved into the handler as it's AWS specific?

lambdas/services/bulk_upload_report_service.py
NogaNHS, line 63: maybe raise BulkUploadReportException here and handle the response in the lambda handler

joefong-nhs (approved):
Looks fine to me. Tested in sandbox that the report lambda can create report and save to S3 bucket as expected.

PR #174, opened 2023-11-30
PRMDR-470: Refactor back channel logout handler

NogaNHS:


lambdas/services/back_channel_logout_service.py
```diff
@@ -0,0 +1,65 @@
...
+            logger.error(f"Error logging out user: {e}")
+            raise LogoutFailureException("Internal error logging user out")
+
+    def finding_session_id_by_sid(self, sid):
+        filter_sid = Attr("sid").eq(sid)
+        db_response = self.__class__.dynamodb_service.scan_table(
+            table_name=self.dynamodb_name, filter_expression=filter_sid
+        )
+        items = db_response.get("Items", None)
+        print(type(items))
```
joefong-nhs, line 51: this print statement probably is some leftover stuff that we can remove?

NogaNHS: Yes 😅

joefong-nhs (approved):

PR #178, opened 2023-12-01
PRMDR-475 delete_document_reference_handler: move business logic to service

joefong-nhs:


lambdas/tests/unit/handlers/test_delete_document_reference_handler.py
NogaNHS, line 1: Can you add a test for Service throwing an error

lambdas/services/document_service.py
```diff
@@ -121,3 +121,21 @@ def delete_documents(
...
+        document_references: list[DocumentReference],
+        type_of_delete: str,
+    ):
+        table_name = ""
+        if doc_type == SupportedDocumentTypes.ARF.value:
+            table_name = os.environ["DOCUMENT_STORE_DYNAMODB_NAME"]
+        elif doc_type == SupportedDocumentTypes.LG.value:
+            table_name = os.environ["LLOYD_GEORGE_DYNAMODB_NAME"]
+
+        return self.delete_documents(
```
NogaNHS, line 137: If table name is empty we shouldn't call this.

lambdas/tests/unit/handlers/test_delete_document_reference_handler.py
joefong-nhs: Done :D  Thank you.

lambdas/services/document_service.py
joefong-nhs: In the other methods in DocumentService, I don't see a pattern of special handling when table_name or doc_type is empty. 
When we run delete_document with table_name empty, boto3 would raise a client error, which will be handled by the lambda handler.

lambdas/handlers/delete_document_reference_handler.py
```diff
@@ -29,74 +26,36 @@
...
             )
+            return ApiGatewayResponse(
+                404, "No documents available", "DELETE"
+            ).create_api_gateway_response()
 
     except ClientError as e:
         logger.info(str(e), {"Result": f"Unsuccessful deletion due to {str(e)}"})
         return ApiGatewayResponse(
             500, "Failed to delete documents", "DELETE"
         ).create_api_gateway_response()
```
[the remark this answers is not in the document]
joefong-nhs: Thanks :) 
I grabbed the LambdaException class from your branch and added a decorator to catch that. (and added unit tests around that) 
I am still keeping this ClientError except block as a safety net for AWS-related errors, but if any underlying code raises our custom LambdaException it will be handled now.

lambdas/services/document_service.py
```diff
@@ -121,3 +121,21 @@ def delete_documents(
...
+        self,
+        doc_type: str,
+        document_references: list[DocumentReference],
+        type_of_delete: str,
+    ):
+        table_name = ""
+        if doc_type == SupportedDocumentTypes.ARF.value:
+            table_name = os.environ["DOCUMENT_STORE_DYNAMODB_NAME"]
+        elif doc_type == SupportedDocumentTypes.LG.value:
+            table_name = os.environ["LLOYD_GEORGE_DYNAMODB_NAME"]
```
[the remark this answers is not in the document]
NogaNHS: I understand, but I don't think this service should handle getting env vars. Maybe we can discuss this with the team?

lambdas/services/document_service.py
NogaNHS: I don't think we should get to boto3 with invalid request, even if it would raise ClientError. If the other methods in the service has the same issue I assume it would be refactor as part of another ticket?

joefong-nhs: I have changed the implementation, now table_name shouldn't be empty in any case.

lambdas/enums/supported_document_types.py
```diff
@@ -20,3 +26,42 @@ def get_from_field_name(value: str):
...
+            SupportedDocumentTypes.ARF.get_dynamodb_table_name()
+            (returns "ndr*_DocumentReferenceMetadata")
+
+        result:
+            "ndr*_DocumentReferenceMetadata"
+
+        Eventually we could replace all os.environ["XXX_DYNAMODB_NAME"] calls with this method,
+        so that the logic of resolving table names are controlled in one place.
+        """
+        try:
```
abbas-khan10, line 44: For this function do you think it might be sensible to return an array of tablenames (in the case of SupportedDocumentTypes.ALL), this way in DocumentService fetch by doc type we can do a for each on the array of tablenames and query this way now instead of the if statement block that exists currently

joefong-nhs: I quite like this idea! :D  Just implemented it and added a unit test.

lambdas/utils/decorators/handle_lambda_exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+    Usage:
+    @handle_lambda_exceptions
+    def lambda_handler(event, context):
+        ...
+    """
+
+    def interceptor(event, context):
+        try:
+            return lambda_func(event, context)
+        except LambdaException as e:
```
NogaNHS, line 20: Where are we raising LambdaException? I thought we're just going to use it as a base class

lambdas/tests/unit/services/test_document_deletion_service.py
```diff
@@ -0,0 +1,164 @@
...
+    ["doc_type", "table_name", "doc_ref"],
+    [
+        (SupportedDocumentTypes.ARF, MOCK_ARF_TABLE_NAME, TEST_DOC_STORE_REFERENCES),
+        (SupportedDocumentTypes.LG, MOCK_LG_TABLE_NAME, TEST_LG_DOC_STORE_REFERENCES),
+    ],
+)
+def test_delete_specific_doc_type(
+    set_env, doc_type, table_name, doc_ref, mocker, mock_document_query
+):
+    service = DocumentDeletionService()
```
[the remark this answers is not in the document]
joefong-nhs: Thanks! Have moved those into fixtures.

PR #180, opened 2023-12-04
PRMDR-474: Refactor to Create Doc Ref lambda

NogaNHS:


lambdas/services/create_document_reference_service.py
```diff
@@ -0,0 +1,131 @@
...
+from utils.audit_logging_setup import LoggingService
+from utils.exceptions import CreateDocumentRefException, InvalidResourceIdException
+from utils.lloyd_george_validator import validate_lg_files, LGInvalidFilesException
+from utils.utilities import create_reference_id, validate_id
+
+logger = LoggingService(__name__)
+
+
+class CreateDocumentReferenceService:
+    def __init__(self, nhs_number, s3_service_class, dynamo_service_class):
```
abbas-khan10, line 17: Did we agree to set the base aws services as class variables for our custom services? Seems you're not really using them in the handler but creating them and then passing to this service

lambdas/handlers/create_document_reference_handler.py
```diff
@@ -41,141 +34,48 @@ def lambda_handler(event, context):
...
-        if document_type == SupportedDocumentTypes.LG.value:
-            document_reference = NHSDocumentReference(
-                nhs_number=nhs_number,
-                s3_bucket_name=lg_s3_bucket_name,
-                reference_id=s3_object_key,
-                content_type=document.contentType,
-                file_name=document.fileName,
+        if not nhs_number:
+            raise CreateDocumentRefException(
+                400, "Request body missing some properties"
```
joefong-nhs, line 69: Instead of calling multiple .get(), I think we could just try to drill down the nested object and catch all KeyError & TypeError. That would give better readability and is more aligned with the EAFP principle of python.

For example we could do it like:

```
def processing_event_details(event):
    try:
        body = json.loads(event["body"])

        if not body or not isinstance(body, dict):
            raise CreateDocumentRefException(400, "Missing request body")

        nhs_number = body["subject"]["identifier"]["value"]
        content = body["content"][0]
        doc_list = content["attachment"]

        return nhs_number, doc_list

    except (JSONDecodeError, AttributeError) as e:
        logger.error(e)
        raise CreateDocumentRefException(400, f"Invalid json in body: {str(e)}")
    except (KeyError, TypeError) as e:
        logger.error(e)
        raise CreateDocumentRefException(400, "Request body missing some properties")
```

abbas-khan10 (approved):

lambdas/handlers/create_document_reference_handler.py
NogaNHS: Done

joefong-nhs (commented):
I have left a few comments mainly related to create_doc_ref service. Happy to discuss in tech huddle or on slack.

Seems this is the first time we use dynamo batch writer, so I think we've better deploy this to a sandbox and manually call the create doc ref endpoint, just so to confirm boto3 batch writer work as we expected.
Also, suggest to run `make format` before we merge

lambdas/services/create_document_reference_service.py
```diff
@@ -0,0 +1,131 @@
...
+            validated_doc: UploadRequestDocument = UploadRequestDocument.model_validate(
+                document
+            )
+        except ValidationError:
+            raise CreateDocumentRefException(
+                400, "Failed to parse document upload request data"
+            )
+
+        document_type = SupportedDocumentTypes.get_from_field_name(
+            validated_doc.docType
```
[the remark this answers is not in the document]
NogaNHS: Not sure I understand where you're getting to here. We still need self.lg_document to validate lg.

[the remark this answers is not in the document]
NogaNHS: Also we can have both LG and ARF type docs in one request

PR #188, opened 2023-12-06
PRMDR-480: Refactor patient search

NogaNHS:


lambdas/tests/unit/handlers/test_search_patient_details_handler.py
```diff
@@ -18,126 +17,77 @@ def patch_env_vars():
...
+def mocked_context(mocker):
+    mocked_context = mocker.MagicMock()
+    mocked_context.authorization = {
+        "selected_organisation": {"org_ods_code": "Y12345"},
+        "repository_role": "GP_ADMIN",
+    }
+    _ = mocker.patch(
+        "handlers.search_patient_details_handler.request_context", mocked_context
+    )
+    yield _
```
joefong-nhs, line 30: I think here we could yield the patched object directly

```suggestion
    yield mocker.patch(
        "handlers.search_patient_details_handler.request_context", mocked_context
    )
```

lambdas/handlers/search_patient_details_handler.py
```diff
@@ -27,89 +21,21 @@ def lambda_handler(event, context):
     try:
         nhs_number = event["queryStringParameters"]["patientId"]
         request_context.patient_nhs_no = nhs_number
+        user_ods_code, user_role = "", ""
+        if isinstance(request_context.authorization, dict):
+            user_ods_code = request_context.authorization.get(
+                "selected_organisation", {}
+            ).get("org_ods_code", "")
+            user_role = request_context.authorization.get("repository_role", "")
+        search_service = SearchPatientDetailsService(user_ods_code, user_role)
```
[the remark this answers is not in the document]
joefong-nhs: I think line 30~line 31 would handle the case if one of them is empty, so no need a double check before initialising the service.

lambdas/tests/unit/handlers/test_search_patient_details_handler.py
```diff
@@ -18,126 +17,77 @@ def patch_env_vars():
...
 def test_lambda_handler_valid_id_returns_200(
-    valid_id_event_with_auth_header, context, mocker, patch_env_vars
+    valid_id_event_with_auth_header, context, mocker, mocked_context
 ):
-    response = Response()
-    response.status_code = 200
-    response._content = json.dumps(PDS_PATIENT).encode("utf-8")
+    patient_details = """{"givenName":["Jane"],"familyName":"Smith","birthDate":"2010-10-22",
+        "postalCode":"LS1 6AE","nhsNumber":"9000000009","superseded":false,
+        "restricted":false,"generalPracticeOds":"Y12345","active":true}"""
```
joefong-nhs, line 37: Just a trivial thing, this json string of patient detail probably is the same as the one at `test_search_patient_details_service.py` line 82, so I think we might put this string in conftest file so that its easier to reuse in the future.

But its up to you and I am happy to keep it as it is now :)

joefong-nhs (approved):
Nice one.  Gave some little suggestions but I think it's good as it is now.

PR #189, opened 2023-12-06
PRMDR-472 bulk_upload_metadata_handler: move business logic to service

joefong-nhs:


lambdas/services/bulk_upload_metadata_service.py
```diff
@@ -0,0 +1,85 @@
...
+        self.clear_temp_storage()
+
+    def download_metadata_from_s3(self, metadata_filename: str) -> str:
+        local_file_path = os.path.join(self.temp_download_dir, metadata_filename)
+        self.s3_service.download_file(
+            s3_bucket_name=self.staging_bucket_name,
+            file_key=metadata_filename,
+            download_path=local_file_path,
+        )
+        return local_file_path
```
NogaNHS, line 48: possibly client error here?

lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
```diff
@@ -0,0 +1,241 @@
...
+            message_body=EXPECTED_SQS_MSG_FOR_PATIENT_1234567890,
+            nhs_number="1234567890",
+        ),
+        call(
+            group_id="bulk_upload_123412342",
+            queue_url=MOCK_LG_METADATA_SQS_QUEUE,
+            message_body=EXPECTED_SQS_MSG_FOR_PATIENT_1234567891,
+            nhs_number="1234567891",
+        ),
+    ]
```
NogaNHS, line 48: Can we move this expected up to have the test in a Arrange-Act-Assert pattern?

lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
```diff
@@ -0,0 +1,241 @@
...
+        {"Error": {"Code": "500", "Message": "file not exist in bucket"}},
+        "s3_get_object",
+    )
+
+    with pytest.raises(ClientError):
+        service.download_metadata_from_s3(metadata_filename)
+
+
+def test_csv_to_staging_metadata(set_env):
+    service = BulkUploadMetadataService()
```
NogaNHS, line 144: Maybe move the service to a fixture?

lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def test_send_metadata_to_sqs(set_env, mocker, mock_sqs_service):
+    service = BulkUploadMetadataService()
+    mock_parsed_metadata = EXPECTED_PARSED_METADATA
+    mocker.patch("uuid.uuid4", return_value="123412342")
+
+    service.send_metadata_to_fifo_sqs(mock_parsed_metadata)
+
+    expected_calls = [
```
NogaNHS, line 164: Arrange-Act-Assert here too

lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
joefong-nhs: Of course!  Done just now :D

lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
joefong-nhs: Done this as well. Thank :)

lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
joefong-nhs: Agree. I have replaced all these calls with a `metadata_service` fixture.

lambdas/handlers/bulk_upload_metadata_handler.py
```diff
@@ -53,52 +33,3 @@ def lambda_handler(_event, _context):
                 f'No metadata file could be found with the name "{METADATA_FILENAME}"',
                 {"Result": "Unsuccessful bulk upload"},
```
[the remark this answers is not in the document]
joefong-nhs: Just that the control flow would be a bit simpler.
If we catch the error within service, we would need to either throw again or put an explicit return statement to stop the process.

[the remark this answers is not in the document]
joefong-nhs: Had a chat with Noga and now I agreed with Noga's explanation for catching error in service class. 
I have updated the code and tests accordingly.

PR #190, opened 2023-12-06
PRMDR-485 Smoke test workflow

RioKnightleyNHS:


joefong-nhs (approved):
Code looks all-right. Checked that cypress tests are all green when run locally.  

Saw that rio did a test run with the amended github action which is all green as well. image

PR #191, opened 2023-12-06
[PRMDR-471]

AlexHerbertNHS:
Refactoring Bulk Upload Service


lambdas/services/bulk_upload_service.py
```diff
@@ -3,61 +3,78 @@
...
+                logger.info(f"Processing message {index} of {len(records)}")
+                self.handle_sqs_message(message)
+            except PdsTooManyRequestsException as error:
+                logger.error(error)
+                logger.info("Cannot process for now due to PDS rate limit reached.")
+                logger.info(
+                    "All remaining messages in this batch will be returned to sqs queue to retry later."
+                )
+
+                all_unprocessed_message = records[index - 1 :]
```
RachelHowellNHS, line 58:
```suggestion
                all_unprocessed_message = records[index - 1:]
```

lambdas/handlers/bulk_upload_handler.py
```diff
@@ -13,35 +12,27 @@
 @override_error_check
 def lambda_handler(event, _context):
     logger.info("Received event. Starting bulk upload process")
+
+    http_status_code = 200
+    response_body = f"Finished processing all {len(event['Records'])} messages"
```
joefong-nhs, line 17: Maybe could we move these two lines to between lines 29 and 30 for clarity?  So that it is easier to see which status code and response body are related to which scenario.

lambdas/utils/lloyd_george_validator.py
```diff
@@ -167,7 +173,7 @@ def validate_with_pds_service(file_name_list: list[str], nhs_number: str):
         if patient_details.general_practice_ods != current_user_ods:
             raise LGInvalidFilesException("User is not allowed to access patient")
 
-    except (ValidationError, ClientError, ValueError) as e:
+    except (ClientError, ValueError) as e:
```
joefong-nhs, line 176: Would it be safer to keep ValidationError in this catch block?  
Although the chance is slim, but I think ValidationError may occur at line 156 when we try to parse PDS's response

lambdas/services/bulk_upload_service.py
```diff
@@ -3,61 +3,78 @@
...
+                for unprocessed_message in all_unprocessed_message:
+                    self.sqs_repository.put_sqs_message_back_to_queue(
+                        unprocessed_message
+                    )
+                return
```
joefong-nhs, line 63: would it be good if we raise a BulkUploadException here instead of the return statement?

RachelHowellNHS (approved):

PR #193, opened 2023-12-07
PRMDR-476 - Refactor document manifest

abbas-khan10:


lambdas/services/document_manifest_service.py
```diff
@@ -6,33 +6,56 @@
...
-        self.documents = documents
-        self.zip_output_bucket = zip_output_bucket
-        self.zip_trace_table = zip_trace_table
+        self.documents: list[DocumentReference] = []
         self.zip_file_name = f"patient-record-{self.nhs_number}.zip"
         self.temp_downloads_dir = tempfile.mkdtemp()
         self.temp_output_dir = tempfile.mkdtemp()
+        self.zip_output_bucket = os.environ["ZIPPED_STORE_BUCKET_NAME"]
+        self.zip_trace_table = os.environ["ZIPPED_STORE_DYNAMODB_NAME"]
+        # TODO - Add time to live for zipped manifest
```
AlexHerbertNHS, line 33: Is this comment needed? SonarLint doesn't like Todo's, and if this is a ticket later, surely it'd be simple to add this back in?

lambdas/handlers/document_manifest_by_nhs_number_handler.py
```diff
@@ -33,68 +29,34 @@
 @override_error_check
 def lambda_handler(event, context):
```
AlexHerbertNHS, line 30: @ensure_environment_variables might be good to add

lambdas/tests/unit/services/test_document_service.py
AlexHerbertNHS, line 1: SonarCloud is reporting a lot of duplicate lines in this file, not sure this is worth a look given the time constraint but I thought it would be worth pointing out

lambdas/handlers/document_manifest_by_nhs_number_handler.py
```diff
@@ -33,68 +29,34 @@
...
     request_context.app_interaction = LoggingAppInteraction.DOWNLOAD_RECORD.value
-
-    logger.info("Starting document manifest process")
```
AlexHerbertNHS, line 37: Our best practices doc asks us to include an entry point log so we can check the handler is reachable. I feel this should go back in, or something a bit more explicit  about what's running if you're feeling spicy

lambdas/handlers/document_manifest_by_nhs_number_handler.py
```diff
@@ -33,68 +29,34 @@
...
-                204, "No documents found for given NHS number and document type", "GET"
-            ).create_api_gateway_response()
-
         logger.info("Starting document manifest process")
-        document_manifest_service = DocumentManifestService(
-            nhs_number=nhs_number,
-            documents=documents,
-            zip_output_bucket=zip_output_bucket,
-            zip_trace_table=zip_trace_table_name,
+        document_manifest_service = DocumentManifestService(nhs_number)
```
AlexHerbertNHS, line 38: Not sre NHS number should be a part of the object init, I don't see the value added by having this as a class level parameter vs adding it in as a param to the method call since it's only used once to get the docs list from Dynamo

lambdas/services/document_manifest_service.py
```diff
@@ -6,33 +6,56 @@
...
+    def __init__(self, nhs_number):
         self.s3_service = S3Service()
         self.dynamo_service = DynamoDBService()
+        self.document_service = DocumentService()
+
         self.nhs_number = nhs_number
-        self.documents = documents
-        self.zip_output_bucket = zip_output_bucket
-        self.zip_trace_table = zip_trace_table
+        self.documents: list[DocumentReference] = []
```
AlexHerbertNHS, line 27: I don't feel this should be object level

lambdas/services/document_manifest_service.py
```diff
@@ -6,33 +6,56 @@
...
         self.zip_file_name = f"patient-record-{self.nhs_number}.zip"
```
AlexHerbertNHS, line 28: As above, I think we should make this as it's needed and not in the init

AlexHerbertNHS (changes requested):
Generally looks good, happy to discuss the requested changes if you feel they're not relevant

lambdas/handlers/document_manifest_by_nhs_number_handler.py
abbas-khan10: It's already in if you expand the code up it's just hidden from view

lambdas/services/document_service.py
```diff
@@ -12,10 +12,10 @@
...
 
-class DocumentService(DynamoDBService):
+class DocumentService:
     def __init__(self):
-        super().__init__()
         self.s3_service = S3Service()
+        self.dynamo_service = DynamoDBService()
 
     def fetch_available_document_references_by_type(
         self, nhs_number: str, doc_type: str
```
AlexHerbertNHS, line 21: on line 23 there's a TODO in this file still - can we nic that or implement it?

lambdas/services/document_manifest_service.py
```diff
@@ -71,21 +94,21 @@ def download_documents_to_be_zipped(self):
...
+                    status_code=500,
+                    message=f"Reference to {document.file_key} that doesn't exist in s3",
                 )
 
     def upload_zip_file(self):
         logger.info("Creating zip from files")
-        temp_output_dir = tempfile.mkdtemp()
 
-        zip_file_path = os.path.join(temp_output_dir, self.zip_file_name)
+        zip_file_path = os.path.join(self.temp_output_dir, self.zip_file_name)
```
AlexHerbertNHS, line 105: Also this method is the only one that uses Dynamo Service - I feel that since the document_service below contains all the logic for getting the DynamoDB info for this use case (I'd have called this , we should maybe move the last line of this method into that service?

lambdas/handlers/document_manifest_by_nhs_number_handler.py
abbas-khan10: It was removed due to duplication on line 37

lambdas/handlers/document_manifest_by_nhs_number_handler.py
abbas-khan10: It's used to generate the zip file name upon initialisation, which is used across all class methods so I think it seems sensible to keep in the init

lambdas/services/document_manifest_service.py
abbas-khan10: It's shared across multiple class methods so the alternative would be to pass it in as a parameter to said class methods, with that being the case keeping it at object level so it's methods have access to it seems sensible

lambdas/services/document_manifest_service.py
abbas-khan10: Similar here, it's passed to the s3 service in 2 separate methods so seems sensible to keep as an object variable, otherwise we'd just be duplicating the same string

lambdas/services/document_manifest_service.py
abbas-khan10: I agree, this one can be removed as it's only really used in one function

lambdas/services/document_manifest_service.py
abbas-khan10: No just a reminder that we need to add TTL to the zip process at some point but written a note down and removed

lambdas/tests/unit/services/test_document_service.py
abbas-khan10: Yeah there's going to be a few test files which will complain, might need to leave that for future refactoring and properly implemented fixtures

lambdas/handlers/document_manifest_by_nhs_number_handler.py
AlexHerbertNHS: bwahaha I pulled it and was looking but missed it anyways! What a n00b

PR #198, opened 2023-12-11
PRMDR-413 refactor registration validation for gp admin role

RachelHowellNHS:


lambdas/tests/unit/handlers/test_token_handler.py
```diff
@@ -76,4 +73,46 @@ def test_handler_passes_error_details_in_response(
 
     assert actual == expected
 
-    mock_login_service["exchange_token"].expect_called_with(auth_code, state)
+    mock_login_service.generate_session.assert_called_with(state, auth_code)
+
+
+def test_missing_query_string_params_raise_key_error(mock_login_service, context):
```
thisusernameisnowtaken, line 79: Thank you for adding these, this was an oversight from the refactoring ticket

PR #200, opened 2023-12-13
PRMDR-500: Python package check and add pre-commit hooks for lambdas

abbas-khan10:


joefong-nhs (approved):
Tested in local that the pre-commit hook did trigger python formatter. Also github actions seems working well. 
I think its all good to go :)

PR #201, opened 2023-12-13
PRMDR-464 Remove hardcoded port numbers

joefong-nhs:
- Remove hardcoded port numbers 80 and 8080 (at `nginx.conf` and `docker-compose.yml`). Replace by reading the port number from AWS parameter store or local .env file instead.
- Remove hardcoded port number 3000 from multiple file. Replace by reading from github variable or local.env file instead.
- Update `README.md` and `.env.template` to reflect the changes

**Note:** This PR has pulled in the updated cypress tests from another ticket PRMDR-485, so better not to merge this until  PRMDR-485 is merged to main.


AlexHerbertNHS (approved):

app/cypress.config.ts
```diff
@@ -1,4 +1,10 @@
 import { defineConfig } from 'cypress';
+import * as dotenv from 'dotenv';
+
+function getBaseUrlFromEnv() {
+    dotenv.config();
+    return process.env.CYPRESS_BASE_URL;
+}
```
carlsmith101, line 7: pretty sure this shouldn't be necessary - cypress will already automatically populate the baseUrl when CYPRESS_BASE_URL is set. See [here](https://docs.cypress.io/guides/guides/environment-variables#We-can-move-this-into-a-Cypress-environment-variable)

app/cypress/e2e/0-ndr-core-tests/auth_routes/auth_gp_admin_path_access.cy.js
```diff
@@ -10,48 +10,49 @@ const patient = {
     active: true,
 };
 
-const smokeTest = Cypress.env('CYPRESS_RUN_AS_SMOKETEST') ?? false;
-const baseUrl = Cypress.env('CYPRESS_BASE_URL') ?? 'http://localhost:3000/';
+const baseUrl = Cypress.config('baseUrl');
```
carlsmith101, line 13: cypress will automatically populate the baseUrl if this is set in config, so we can replace all instances of cypress.visit(baseUrl + '/search') or similar with simply cypress.visit('/search').

this would be a much cleaner solution

carlsmith101 (changes requested):
see comments

app/cypress.config.ts
joefong-nhs: You are right. This part is mainly for reading the CYPRESS_BASE_URL from a local `.env` file. (dotenv.config() does that), so that we don't need to manually set the CYPRESS_BASE_URL when running in local.

app/cypress/e2e/0-ndr-core-tests/auth_routes/auth_gp_admin_path_access.cy.js
joefong-nhs: Yes, I think Rio has already replaced `cy.visit(baseUrl + '/route')` by `cy.visit('/route')` in most parts 😄  
I did a search just now and fixed one remaining  at e2e.ts, so we should be free from that now.

For the baseUrl in this file, I think is not for cy.visit but for path comparison  `cy.url().should('eq', baseUrl + '/search/upload');`, which probably still needs the baseUrl provided

carlsmith101 (approved):

PR #204, opened 2023-12-19
PRMDR-488 - Add test data provisioning cypress commands

carlsmith101:


app/cypress/support/aws.config.ts
```diff
@@ -0,0 +1,11 @@
+import AWS from 'aws-sdk';
+
+// need to get from env vars
+AWS.config.update({
```
carlsmith101, line 4: This needs to be updated to pass in these values from environment variables. Need to clarify approach to follow existing implementations

RachelHowellNHS: Is this to be left as is and updated in a future ticket?

RachelHowellNHS (approved):
LGTM :) left a comment but assume the answer is yes, in which case all good

PR #213, opened 2024-01-05
PRMDR-549 - Update start page UI with new content (v1.1)

joefong-nhs:


app/src/pages/homePage/HomePage.test.tsx
```diff
@@ -28,29 +29,42 @@ describe('StartPage', () => {
...
+
         contentStrings.forEach((s) => {
             expect(screen.getByText(s)).toBeInTheDocument();
         });
+        Object.entries(contentStringsAppearMoreThanOnce).forEach(([s, count]) => {
+            const foundElements = screen.getAllByText(s);
+            expect(foundElements).toHaveLength(count);
+            foundElements.forEach((element) => expect(element).toBeInTheDocument());
+        });
+
```
RioKnightleyNHS, line 59: This feels like over-engineering to check for multiple instances of a string only once. Maybe if there was a a bigger map.

` 
const foundElements = screen.getAllByText('download a patient record');
expect(foundElements).toHaveLength(2);
`
This should be fine

RioKnightleyNHS (changes requested):

app/src/pages/homePage/HomePage.test.tsx
joefong-nhs: Agree. Probably we won't have another duplicated string on this page. Thanks and suggestion adopted :)

RioKnightleyNHS (approved):

PR #217, opened 2024-01-05
PRMDR-558: Multiple GP admin role codes

abbas-khan10:


lambdas/utils/constants/ssm.py
```diff
@@ -0,0 +1,6 @@
+GP_ADMIN_USER_ROLE_CODE = "/auth/smartcard/role/gp_admin_multiple"
```
SRAlexander, line 1: Is it worth naming GP_ADMIN_USER_ROLE_CODES?

lambdas/services/token_handler_ssm_service.py
```diff
@@ -14,68 +20,116 @@ def get_smartcard_role_codes(self) -> list[str]:
...
-        response = params["/auth/smartcard/role/gp_admin"]
+        params = self.get_ssm_parameters([GP_ADMIN_USER_ROLE_CODE])
+        values = params.get(GP_ADMIN_USER_ROLE_CODE)
+
+        if values is None:
+            logger.error(
+                "SSM parameter values for GP admin role may not exist",
+                {"Result": "Unsuccessful login"},
+            )
+            raise LoginException(
```
SRAlexander, line 58: Tempted to say this should not be an explicit LoginException, this is more a SSMException

SRAlexander (changes requested):

lambdas/services/token_handler_ssm_service.py
abbas-khan10: I considered this, the reason it's a LoginException is it's an extension of the LambdaException class and not a regular exception so will be caught as part of the lambda exception handler decorator. Also, since this token ssm service is only used by the token handler, made sense to make it a LoginException instead. The alternative would be creating a new SSMException, passing it up to the login service, catching it there and then throwing the LoginException up to the handler so thought it'd be easier to cut out the middleman, but can do either really

lambdas/utils/constants/ssm.py
abbas-khan10: Potentially, I guess to be clear it's more of an ssm path but GP_ADMIN_USER_ROLE_CODE_SSM_PATH, seemed a bit too long. Can change this if it's not clear

lambdas/services/token_handler_ssm_service.py
SRAlexander: Okay makes sense, let's monitor this one going forwards. We may need a SSMException in other Lambdas going forwards, if we do, let's hot swap it then.

SRAlexander (approved):

PR #219, opened 2024-01-08
deleteAfterDownload added for download page and configured into LG Re…

SRAlexander:
…cord page


SRAlexander:
Tests are more suitable to Cypress work journeys so will be adding on that specific ticket

abbas-khan10 (approved):

PR #234, opened 2024-01-11
Prmdr 394: Inform user if call to document manifest fails

thisusernameisnowtaken:


RioKnightleyNHS (approved):

app/src/pages/documentSearchResultsPage/DocumentSearchResultsPage.tsx
```diff
@@ -107,6 +108,15 @@ function DocumentSearchResultsPage() {
                 </>
             )}
 
+            {(submissionState === SUBMISSION_STATE.FAILED ||
+                downloadState === SUBMISSION_STATE.FAILED) && (
+                <ErrorBox
```
joefong-nhs, line 113: How about if we just check downloadState === SUBMISSION_STATE.FAILED here?  Because it seems a bit odd to display this message when the search failed and the user didn't click the download button

```suggestion
            {downloadState === SUBMISSION_STATE.FAILED && (
                <ErrorBox
```

thisusernameisnowtaken: I've removed the condition and tested the code

PR #238, opened 2024-01-12
[PRMDR 583] Return a HTTP 401 if a user logs in with an invalid role

thisusernameisnowtaken:


lambdas/tests/unit/services/test_login_service.py
```diff
@@ -356,6 +358,7 @@ def test_generate_repository_role_no_role(mock_logging_service, set_env, mocker)
 
     login_service = LoginService()
 
-    expected = RepositoryRole.NONE
-    actual = login_service.generate_repository_role(org, user_role_code)
-    assert expected == actual
+    with pytest.raises(LoginException) as actual:
+        actual = login_service.generate_repository_role(org, user_role_code)
```
joefong-nhs, line 362: the raises block already assigned a value for `actual` so no need to assign it again
```suggestion
        login_service.generate_repository_role(org, user_role_code)
```

joefong-nhs (approved):

PR #239, opened 2024-01-12
PRMDR-553 update routes and verify patient page text

RachelHowellNHS:


app/src/types/generic/routes.ts
```diff
@@ -16,8 +16,8 @@ export enum routes {
 
     LLOYD_GEORGE = '/search/patient/lloyd-george-record',
 
-    UPLOAD_SEARCH = '/search/upload',
-    UPLOAD_VERIFY = '/search/upload/result',
+    PATIENT_SEARCH = '/search',
+    VERIFY_PATIENT = '/search/result',
     UPLOAD_DOCUMENTS = '/upload/submit',
 }
```
joefong-nhs, line 23: Just a suggestion, but would it be good if we use another word for example "/gp" to replace "/upload" instead of just deleting "/upload"?
Because we already got a `DOWNLOAD_DOCUMENTS = '/search/results'` definition on line 15. 
I think having `/search/results` and `/search/result` at the same time is kind of confusing.


```
    PATIENT_SEARCH = '/search/gp',
    VERIFY_PATIENT = '/search/gp/result',
```

RachelHowellNHS: Good point, I was thinking that because there is a ticket to get rid of the duplicate routes it would resolve this issue but I will check with Rio to see if that's going to be done soon as that ticket isn't in the current sprint

SRAlexander: If anything the Download Documents url sounds a little off too, do we know what page it's used and can we come up with a better route for it?

RioKnightleyNHS:
Awaiting discussed sitemap and cypress changes

RioKnightleyNHS (changes requested):
Awaiting discussed cypress and sitemap changes

RioKnightleyNHS (changes requested):

RioKnightleyNHS (approved):

PR #243, opened 2024-01-12
(UI ONLY) PRMDR-577 - Enable feedback link and create Feedback form UI 

joefong-nhs:
- Added a feedback page with a working input form
- Added a link at the phase banner which opens the feedback page in a new tab

(The functionality of sending email depends on PRMDR-581 so not included yet. For now, submitting the form calls a console.log and display the input data at the bottom of page)


app/src/types/pages/feedbackPage/types.ts
RachelHowellNHS, line 1: Just curious, what is the naming convention for enums, what is the difference between the camel case keys and the pascal case ones?

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+                    <Input
+                        label="Your name"
+                        data-testid={FORM_FIELDS.respondentName}
+                        autoComplete="name"
+                        spellCheck={false}
+                        {...respondentNameProps}
+                    />
+
+                    <Input
```
RachelHowellNHS, line 100: do we need an email type on this input or something else to validate it?

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+} from '../../../types/pages/feedbackPage/types';
+import React, { Dispatch, useState } from 'react';
+import { SubmitHandler, useForm, UseFormRegisterReturn } from 'react-hook-form';
+import sendEmail from '../../../helpers/requests/sendEmail';
+import { Button, Fieldset, Input, Radios, Textarea } from 'nhsuk-react-components';
+import SpinnerButton from '../../generic/spinnerButton/SpinnerButton';
+
+export type Props = {
+    readonly stage: SUBMISSION_STAGE;
+    readonly setStage: Dispatch<SUBMISSION_STAGE>;
```
RioKnightleyNHS, line 15: Props in React are immutable by default, this is an anti-pattern and not necessary

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+    );
+}
+
+const renameRefKey = (props: UseFormRegisterReturn, newRefKey: string) => {
+    const { ref, ...otherProps } = props;
+    return {
+        [newRefKey]: ref,
+        ...otherProps,
+    };
+};
```
RioKnightleyNHS, line 133: Good idea, but can we move the creation of this to where the register callbacks happen. Can create some confusion to how react-hook-forms is working to some who know less UI, if it's at the bottom of the page

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
RioKnightleyNHS: Agree, adding email regex to prevent aws ses client error on submit with invalid email.

There is an example on the react-hook-form website

`<input
        id="email"
        {...register("email", {
          required: "required",
          pattern: {
            value: /\S+@\S+\.\S+/,
            message: "Entered value does not match email format",
          },
        })}
        type="email"
      />`

app/src/components/layout/phaseBanner/PhaseBanner.tsx
```diff
@@ -9,7 +10,11 @@ function PhaseBanner() {
                         New Service
                     </strong>
                     <span className="govuk-phase-banner__text">
-                        Your feedback will help us to improve this service.
+                        Your{' '}
+                        <a href={routes.FEEDBACK} target="_blank" rel="noreferrer">
```
RioKnightleyNHS: Incorrect useage of the anchor href. These should be for external links only as it does not take advantage of the react-router and hard reloads the entire page. React router is capable of detecting what components remain, and what stays, including our context.

Instead use an onClick with e.preventDefault() + useNavigate. or use a Link to="" component. You can find examples of this around the application

app/src/router/AppRouter.tsx
```diff
@@ -62,6 +64,10 @@ export const routeMap: Routes = {
         page: <UnauthorisedPage />,
         type: ROUTE_TYPE.PUBLIC,
     },
+    [FEEDBACK]: {
```
RioKnightleyNHS, line 67: Private route as discussed in tech huddle

app/src/types/pages/feedbackPage/types.ts
```diff
@@ -0,0 +1,23 @@
...
+    Successful,
+    Failure,
+}
+
+export enum SATISFACTION_CHOICES {
+    VerySatisfied = 'Very satisfied',
+    Satisfied = 'Satisfied',
+    Neither = 'Neither satisfied or dissatisfied',
+    Dissatisfied = 'Dissatisfied',
+    VeryDissatisfied = 'Very dissatisfied',
```
RioKnightleyNHS, line 22: Not sure if we should be using enums as strings to render. These are reserved for logical choices within the code. If we were to apply a language library such as i18n we can't use it here. Please extract this into a regular key:value object if you prefer this method for label:value

RioKnightleyNHS (changes requested):

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
joefong-nhs: This is to suppress a so-called "Code Smells" that SonarCloud complains about. I am happy to remove the readonly if its consider an anti-pattern.
 https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=prm-repository_repository&open=AYwfjaMMhuX2Ulp5DlkP

app/src/components/layout/phaseBanner/PhaseBanner.tsx
joefong-nhs: Thanks!!  just replaced the <a> tag with <Link> . Will continue to address other issues as well

app/src/types/pages/feedbackPage/types.ts
joefong-nhs: Thanks. Changed SATISFACTION_CHOICES to a plain key:value object

app/src/router/AppRouter.tsx
joefong-nhs: Change to a private route. 
Also, amend implementation for the phase banner to only display the word "feedback" as a link when logged in

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
joefong-nhs: Thanks both :D  Added email validation (using a validator library)
Now the email address can be either blank or a valid one. If an invalid email address is given, the form will show an alert upon submission.

app/src/types/pages/feedbackPage/types.ts
joefong-nhs: well spotted, just unified both to pascal case :)

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
RachelHowellNHS: I wonder if sonar cloud complained because the setStage prop doesn't have the 'SetStateAction' part we have used elsewhere in the code eg setStage: Dispatch<SetStateAction<LG_RECORD_STAGE>>;
could try adding that and removing readonly to see if it resolves the 'code smell'

app/src/components/blocks/feedbackForm/FeedbackForm.tsx
RioKnightleyNHS: Is email a required field ?

RioKnightleyNHS (approved):

PR #244, opened 2024-01-15
PRMDR 573 Update bulk upload report with ods code

NogaNHS:


SRAlexander (approved):

lambdas/utils/lloyd_george_validator.py
```diff
@@ -177,12 +151,28 @@ def validate_with_pds_service(file_name_list: list[str], nhs_number: str):
     except (ValidationError, ClientError, ValueError) as e:
         logger.error(e)
         raise LGInvalidFilesException(e)
+
+
+def getting_patient_info_from_pds(nhs_number: str):
```
joefong-nhs, line 156: maybe add the return type here

```suggestion
def getting_patient_info_from_pds(nhs_number: str) -> PatientDetails:
```

lambdas/utils/lloyd_george_validator.py
```diff
@@ -131,32 +123,14 @@ def check_for_file_names_agrees_with_each_other(file_name_list: list[str]):
         raise LGInvalidFilesException("File names does not match with each other")
 
 
-def validate_with_pds_service(file_name_list: list[str], nhs_number: str):
+def validate_with_pds_service(
```
joefong-nhs, line 126: would it be good to rename this method to something like `validate_filenames_with_patient_info`?
Because now the functionality of fetching PDS API is moved elsewhere

lambdas/models/bulk_upload_status.py
```diff
@@ -34,6 +35,7 @@ class FailedUpload(UploadStatusBaseClass):
     "Date",
     "Timestamp",
     "ID",
+    "OdsCode",
```
joefong-nhs, line 38: Just a trivial thing, but could we change the order of this list and move "OdsCode" higher? For example after the FailureReason. 
The order of this list will become the column order of report from left to right, so this could help people who follow up the "Patient not registered at your practice" failure

joefong-nhs (approved):
Looks good overall. Gave a few suggestions on minor things, but are good-to-have not must-have

PR #255, opened 2024-01-18
smoke test scenarios for unhappy path

premnhs4:


app/cypress/e2e/1-ndr-smoke-tests/gp_user_workflows/patient_search_and_verify_unhappy_workflow.cy.js
```diff
@@ -0,0 +1,34 @@
...
+                cy.getByTestId('search-patient-btn').should('exist');
+                cy.getByTestId('search-patient-btn').click();
+
+                cy.get('#nhs-number-input').click();
+                cy.get('#nhs-number-input').type(activePatient);
+                cy.get('#search-submit').click();
+                // Assert
+                cy.get('#nhs-number-input--error-message').should(
+                    'include.text',
+                    'Sorry, patient data not found.',
```
joefong-nhs, line 29: There is a visually hidden "Error: " text in that message box (for accessibility reason), so we need to account for that as well

```suggestion
                    'Error: Sorry, patient data not found.',
```

app/cypress/e2e/1-ndr-smoke-tests/gp_user_workflows/patient_search_and_verify_unhappy_workflow.cy.js
```diff
@@ -0,0 +1,34 @@
...
+    const baseUrl = Cypress.config('baseUrl');
+    const gpRoles = [Roles.GP_ADMIN_H85686, Roles.GP_CLINICAL_H85686];
+
+    const workspace = Cypress.env('WORKSPACE');
+    const activePatient = workspace === 'ndr-dev' ? '9730153817' : '9000000002';
+    const homeUrl = '/home';
+
+    gpRoles.forEach((role) => {
+        it(
+            `[Smoke] Show that patient is not accessable for this  ${roleName(role)} `,
```
carlsmith101, line 14: I think the title could be a bit more descriptive, as I am not 100% clear what this test is proving. 

Is the patient inaccessible because the ODS code of the user is different to that of the patient? In which case, a better test name could be 'it does not allow the user to access the record of a patient registered at a different organisation'.

PR #256, opened 2024-01-18
PRMDR-582 Exception Errors

RioKnightleyNHS:


lambdas/utils/error_response.py
```diff
@@ -0,0 +1,244 @@
...
+class ErrorResponse:
+    def __init__(self, err_code: str, message: str) -> None:
+        self.err_code = err_code
+        self.message = message
+
+    def create(self) -> str:
+        return json.dumps({"message": self.message, "err_code": self.err_code})
+
+    def __eq__(self, other):
+        return self.error_code == other.error_code and self.message == other.message
```
joefong-nhs, line 14:
```suggestion
        return self.err_code == other.err_code and self.message == other.message
```

lambdas/handlers/create_document_reference_handler.py
```diff
@@ -63,10 +64,10 @@ def processing_event_details(event):
             f"Invalid json in body: {str(e)}",
```
abbas-khan10, line 64: Might be worth plugging the enum value here also to avoid log messages and exception messages getting out of sync if the messaes values were to change in future. e.g `f"Invalid json in body: {str(e)}"` -> `f"{LambdaError.CreateDocPayload}: {str(e)}"`

lambdas/utils/error_response.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+class LambdaError(Enum):
```
abbas-khan10, line 17: Thinking these enums should probably live in the enums folder?

lambdas/handlers/logout_handler.py
```diff
@@ -40,15 +41,23 @@ def logout_handler(token):
...
-            500, "Error logging user out", "GET"
+            500,
+            ErrorResponse(err_code, msg).create(),
+            "GET",
         ).create_api_gateway_response()
     except (jwt.PyJWTError, KeyError) as e:
         logger.error(
             f"error while decoding JWT: {e}", {"Result": "Unsuccessful logout"}
         )
+        error = LambdaError.LogoutAuth.value
```
abbas-khan10, line 56: Since the error var is created just to pass the `msg` and `err_code` to` ErrorResponse`, is there anyway we can just pass it an error class and the `ErrorResponse` can then extract the `msg` and `err_code` in it's create() method? Would reduce a lot of duplication from what I can see, something like `ErrorResponse(LambdaError.LogoutAuth).create()`

lambdas/handlers/logout_handler.py
```diff
@@ -39,16 +40,21 @@ def logout_handler(token):
...
         return ApiGatewayResponse(
-            500, "Error logging user out", "GET"
+            500,
+            LambdaError.LogoutClient.create_error_body(),
+            "GET",
         ).create_api_gateway_response()
     except (jwt.PyJWTError, KeyError) as e:
         logger.error(
-            f"error while decoding JWT: {e}", {"Result": "Unsuccessful logout"}
+            f"{LambdaError.LogoutAuth.to_str()}: {e}", {"Result": "Unsuccessful logout"}
```
abbas-khan10, line 54: We've replaced "error while decoding jwt" with ""Invalid Authorization header" is the the correct error for this case?

lambdas/handlers/token_handler.py
```diff
@@ -45,6 +42,11 @@ def lambda_handler(event, context):
         ).create_api_gateway_response()
 
     except (KeyError, TypeError):
+        logger.error(
+            f"{ LambdaError.LoginNoAuth.to_str()}", {"Result": "Unsuccessful login"}
```
abbas-khan10, line 46: Could we catch the exception "as e" here and add it to the logger.error?

abbas-khan10 (approved):

PR #269, opened 2024-01-26
PRMDR-608 Privacy Policy

RioKnightleyNHS:


RachelHowellNHS (commented):
On the ticket it says the privacy notice should be opened in a new tab, was that changed? Also, to match the design I think we'd need a max-width on the page?

RachelHowellNHS (approved):

PR #275, opened 2024-01-30
Prmdr-625: E2E test for privacy policy page

premnhs4:


app/cypress/e2e/0-ndr-core-tests/privacy_page.cy.js
```diff
@@ -0,0 +1,62 @@
...
+
+    const loginAndVisitPrivacyPolicyPage = (role) => {
+        cy.login(role);
+        cy.visit(privacypolicyUrl);
+    };
+
+    it('verify privacy policy link in start up page', { tags: 'regression' }, () => {
+        cy.visit(startUrl);
+        cy.url().should('eq', baseUrl + startUrl);
+        cy.get('.nhsuk-footer__list-item-link')
```
SRAlexander, line 17: Is this check needed if we're checking the main link below directly with the data id?

app/cypress/e2e/0-ndr-core-tests/privacy_page.cy.js
```diff
@@ -0,0 +1,62 @@
...
+    rolesToTest.forEach((role) => {
+        const nameOfRole = Roles[role];
+        describe(`Role: ${nameOfRole}`, () => {
+            it(
+                'Opens the privacy-policy page when a user clicks the privacy policy link at the footer',
+                { tags: 'regression' },
+                () => {
+                    cy.login(role);
+
+                    cy.get('.nhsuk-footer__list-item-link')
```
SRAlexander, line 33: Same as above

SRAlexander (approved):

app/cypress/e2e/0-ndr-core-tests/privacy_page.cy.js
```diff
@@ -0,0 +1,62 @@
...
+        cy.visit(startUrl);
+        cy.url().should('eq', baseUrl + startUrl);
+        cy.get('.nhsuk-footer__list-item-link')
+            .should('exist')
+            .and('have.attr', { target: '_blank' });
+        cy.getByTestId('privacy-link').invoke('removeAttr', 'target').click();
+        cy.url().should('eq', baseUrl + privacypolicyUrl);
+    });
+
+    rolesToTest.forEach((role) => {
```
carlsmith101, line 24: it feels like we would benefit from a slightly different approach here. 

We are adding a decent overhead to the total test execution time by running these tests 3 times (once for each role) even though the test assertions do not change dynamically based on each role.

To me, it makes more sense to expand the `auth_routes` tests to include all routes which should be accessible for each role type (including `/privacy-policy` and then include an individual test for the privacy-policy contents if these are required.

carlsmith101 (changes requested):

app/cypress/e2e/0-ndr-core-tests/privacy_page.cy.js
premnhs4: discussed with Dev.. the contents are tested as part of unit tests so checking contents are not really required.

PR #307, opened 2024-02-26
PRMDR-661: LG upload unhappy path

NogaNHS:


app/src/helpers/utils/uploadDocumentValidation.ts
```diff
@@ -0,0 +1,76 @@
...
+    for (let i = 0; i < uploadDocuments.length; i++) {
+        const currentFile = uploadDocuments[i].file;
+        if (currentFile.size > FIVEGB) {
+            errors.push({
+                file: uploadDocuments[i],
+                error: fileUploadErrorMessages.fileSizeError,
+            });
+            continue;
+        }
+        if (currentFile.type !== 'application/pdf') {
```
abbas-khan10, line 27: Is every file we are going to accept always going to be in PDF format?

NogaNHS: Yes, It's only for LG.

app/src/helpers/utils/uploadDocumentValidation.ts
```diff
@@ -0,0 +1,76 @@
...
+    const REGEX_ACCENT_CHARS_IN_NFC = 'À-ž';
+    const REGEX_PATIENT_NAME_PATTERN = `[A-Za-z ${REGEX_ACCENT_CHARS_IN_NFC}${REGEX_ACCENT_MARKS_IN_NFD}]+`;
+    const REGEX_NHS_NUMBER_REGEX = '[0-9]{10}';
+    const REGEX_LLOYD_GEORGE_FILENAME = new RegExp(
+        `[0-9]+of[0-9]+_Lloyd_George_Record_\\[${REGEX_PATIENT_NAME_PATTERN}]_\\[${REGEX_NHS_NUMBER_REGEX}]_\\[\\d\\d-\\d\\d-\\d\\d\\d\\d].pdf`,
+    );
+    const errors: UploadFilesErrors[] = [];
+    const lgFilesNumber = /of[0-9]+/;
+    const FIVEGB = 5 * Math.pow(1024, 3);
+    for (let i = 0; i < uploadDocuments.length; i++) {
```
RachelHowellNHS, line 18: Could you use a for each or for of loop here instead? Also, could the errors.push block be extracted out to a function as it is repeated a few times

app/src/helpers/utils/uploadDocumentValidation.ts
```diff
@@ -0,0 +1,76 @@
...
+        const currentFile = uploadDocuments[i].file;
+        if (currentFile.size > FIVEGB) {
+            errors.push({
+                file: uploadDocuments[i],
+                error: fileUploadErrorMessages.fileSizeError,
+            });
+            continue;
```
RachelHowellNHS, line 25: I think because you have multiple if statements, each will be checked even if the first condition is met so you don't need continue?

app/src/components/blocks/lloydGeorgeFileInputStage/LloydGeorgeFileInputStage.tsx
```diff
@@ -40,15 +42,22 @@ function LloydGeorgeFileInputStage({ documents, setDocuments, setStage }: Props)
...
     const baseHeaders = useBaseAPIHeaders();
-
     const uploadDocuments = async () => {
+        setShowNoFilesMessage(!hasFileInput);
+        if (!hasFileInput) {
+            window.scrollTo(0, 0);
+            return;
+        }
+        setUploadFilesErrors(uploadDocumentValidation(documents));
+        if (!!uploadFilesErrors.length) {
```
RachelHowellNHS, line 57:
```suggestion
        if (uploadFilesErrors.length) {
```
I think this evaluates to true or false already

app/src/components/layout/errorBox/ErrorBox.tsx
```diff
@@ -44,6 +47,17 @@ const ErrorBox = ({
                             </ErrorSummary.Item>
                         )}
                         {messageBody && <p>{messageBody}</p>}
+                        {errorMessageList &&
+                            errorMessageList.map((errorItem) => {
```
RachelHowellNHS, line 51: sonar cloud says it would prefer just errorMessageList?.map instead of errorMessageList && errorMessageList.map

app/src/helpers/utils/uploadDocumentValidation.ts
RachelHowellNHS: misunderstood what this was doing, ignore

app/src/components/blocks/lloydGeorgeFileInputStage/LloydGeorgeFileInputStage.tsx
NogaNHS: uploadFilesErrors.length return a number

RachelHowellNHS: the if statement converts the condition to a boolean already

app/src/helpers/utils/uploadDocumentValidation.ts
NogaNHS: done

RachelHowellNHS (approved):

PR #308, opened 2024-02-28
PRMDR_687 - Updated createDocumentReference to push to staging bucket

SRAlexander:


lambdas/services/create_document_reference_service.py
```diff
@@ -11,108 +11,136 @@
...
+                (document_reference, doc_type) = self.prepare_doc_object(
+                    nhs_number, document
+                )
+                self.check_valid_doc_type(doc_type)
+
+                if doc_type == SupportedDocumentTypes.ARF.value:
+                    arf_documents.append(document_reference)
+                    arf_documents_dict_format.append(document_reference.to_dict())
+
+                if doc_type == SupportedDocumentTypes.LG.value:
```
NogaNHS, line 58: if else here? no need to check for both as you can only type for a doc

lambdas/services/create_document_reference_service.py
```diff
@@ -11,108 +11,136 @@
...
+            validate_nhs_number(nhs_number)
             for document in documents_list:
-                document_reference = self.prepare_doc_object(document)
-                self.prepare_pre_signed_url(document_reference)
-            if self.lg_documents:
-                validate_lg_files(self.lg_documents, self.nhs_number)
+                (document_reference, doc_type) = self.prepare_doc_object(
+                    nhs_number, document
+                )
+                self.check_valid_doc_type(doc_type)
```
NogaNHS, line 52: Is this method necessary as we're checking the types again on lines 54-58?

lambdas/services/create_document_reference_service.py
```diff
@@ -11,108 +11,136 @@
...
             raise CreateDocumentRefException(400, LambdaError.CreateDocNoType)
 
-        logger.info("Provided document is supported")
-
-        (
-            s3_destination,
-            documents_type_list,
-            documents_list_dict_format,
-        ) = self.return_info_by_doc_type(document_type)
+        logger.info(PROVIDED_DOCUMENT_SUPPORTED_MESSAGE)
```
NogaNHS, line 112: Should this be moved as we're not checking the type here anymore?

lambdas/services/create_document_reference_service.py
```diff
@@ -11,108 +11,136 @@
...
+                if doc_type == SupportedDocumentTypes.LG.value:
+                    lg_documents.append(document_reference)
+                    lg_documents_dict_format.append(document_reference.to_dict())
+
+                url_responses[
+                    document_reference.file_name
+                ] = self.prepare_pre_signed_url(document_reference)
+
+            if lg_documents:
+                validate_lg_files(lg_documents)
```
NogaNHS, line 67: This needs the nhs number passed into it.

NogaNHS (commented):
Need rebasing with main

lambdas/services/create_document_reference_service.py
SRAlexander: I'm going with a yes, it's a fail fast condition and reusable logic.

lambdas/services/create_document_reference_service.py
SRAlexander: Added

lambdas/services/create_document_reference_service.py
SRAlexander: I was going to say yes but I've now extracted it into the default section of a switch statement

lambdas/services/create_document_reference_service.py
SRAlexander: Used a switch instead, tidier than else-if

NogaNHS (approved):

PR #324, opened 2024-03-21
PRMDR-738

NogaNHS:


lambdas/services/document_deletion_service.py
```diff
@@ -42,8 +44,12 @@ def delete_specific_doc_type(
         doc_type: Literal[SupportedDocumentTypes.ARF, SupportedDocumentTypes.LG],
     ) -> list[DocumentReference]:
         try:
+            filter_builder = DynamoQueryFilterBuilder()
+
+            filter_builder.add_condition("Deleted", AttributeOperator.EQUAL, "")
```
abbas-khan10, line 49: The filter builder allows method chaining so can use this here: 
`filter_expression =  filter_builder.add_condition("Deleted", AttributeOperator.EQUAL, "").build()`

NogaNHS: Done

SRAlexander (approved):

lambdas/services/document_manifest_service.py
```diff
@@ -36,11 +41,22 @@ def create_document_manifest_presigned_url(
...
+                    nhs_number=self.nhs_number,
+                    doc_type=doc_type,
+                    query_filter=filter_expression,
                 )
             )
+            file_in_progress_message = (
+                "The patients Lloyd George record is in the process of being uploaded"
+            )
+
+            for doc in documents:
```
SRAlexander, line 56: Clean coding perspective, try not to abbreviate. 'document' would be better than 'doc'.

PR #327, opened 2024-03-22
PRMDR-743 - Lambda: Update Create Document Reference Lambda Access Checks

joefong-nhs:


lambdas/utils/common_query_filters.py
```diff
@@ -0,0 +1,29 @@
...
+        attribute=str(DocumentReferenceMetadataFields.DELETED.value),
+        attr_operator=AttributeOperator.EQUAL,
+        filter_value="",
+    )
+    .add_condition(
+        attribute=str(DocumentReferenceMetadataFields.UPLOADED.value),
+        attr_operator=AttributeOperator.EQUAL,
+        filter_value=True,
+    )
+    .set_combination_operator(ConditionOperator.AND)
```
abbas-khan10, line 27: This is a really nice addition to the filter builder, I think Noga added a filter to dynamo_utils.py called `filter_expression_for_available_docs()`, is there any way we can bring that into this file similar to how you've created these common filters for consistency? 

Also the ConditionOperator is set to AND by default so this line isn't necessary unless you were changing to another Operator, but I guess for readability it is more clear to be specific and can be kept in so will leave it up to you

abbas-khan10: Also for this filter do we want to check that Uploaded = True and also Uploading = False? or is this a scenario that cannot happen?

joefong-nhs: Thanks!  The `UploadCompleted` filter was just a placeholder to avoid InternalServerError. 
When I merged with Noga's branch just now, I found that the filter logic is different, so I will prefer `filter_expression_for_available_docs` over UploadCompleted for now (so that I don't accidentally mess up with the logic in that branch)
Later after the story is tested working, we can move filter_expression_for_available_docs() into a common filter AvailableDocs here

SRAlexander (approved):

PR #328, opened 2024-03-22
Prmdr 742 - Retry and State Locking workflow

NogaNHS:


app/src/pages/lloydGeorgeUploadPage/LloydGeorgeUploadPage.tsx
```diff
@@ -148,12 +164,20 @@ function LloydGeorgeUploadPage() {
                     progress: 100,
                 });
             } catch (e) {
+                window.clearInterval(intervalTimer);
```
joefong-nhs, line 167: ``
```suggestion
                window.clearInterval(updateStateInterval);
```

For this line, I think we should call clearInterval with updateStateInterval, not intervalTimer
Because line 144, 145 and line 167 are in the same scope, the new `intervalTimer` state set in line 145 would not be reflected in line 167.

app/src/pages/lloydGeorgeUploadPage/LloydGeorgeUploadPage.test.tsx
```diff
@@ -149,6 +169,8 @@ describe('UploadDocumentsPage', () => {
             mockS3Upload.mockReturnValue(Promise.resolve());
             mockVirusScan.mockReturnValue(DOCUMENT_UPLOAD_STATE.CLEAN);
             mockUploadConfirmation.mockReturnValue(DOCUMENT_UPLOAD_STATE.SUCCEEDED);
+            mockUpdateDocumentState(Promise.resolve());
```
joefong-nhs, line 172:
```suggestion
```
I think we shouldn't have this line, because this is not mocking updateDocumentState, but calling the mocked function directly in the test.
When I removed this line locally, the test still passed, so probably it isn't needed as well.

app/src/helpers/requests/uploadDocument.test.ts
```diff
@@ -0,0 +1,28 @@
+import axios, { AxiosError } from 'axios';
+import { buildDocument, buildTextFile } from '../test/testBuilders';
+import { DOCUMENT_UPLOAD_STATE as documentUploadStates } from '../../types/pages/UploadDocumentsPage/types';
+import { updateDocumentState } from './uploadDocuments';
+
+// Mock out all top level functions, such as get, put, delete and post:
+jest.mock('axios');
+const mockedAxios = axios as jest.Mocked<typeof axios>;
+
+// ...
```
joefong-nhs, line 10:
```suggestion
```
guess this line was from copy & paste. probably don't need this

app/src/pages/lloydGeorgeUploadPage/LloydGeorgeUploadPage.test.tsx
```diff
@@ -60,8 +76,11 @@ jest.mock(
     '../../components/blocks/lloydGeorgeUploadFailedStage/LloydGeorgeUploadFailedStage',
     () => () => <h1>Mock file failed stage</h1>,
 );
+jest.mock('react-router', () => ({
+    useNavigate: () => mockNavigate,
+}));
 
-describe('UploadDocumentsPage', () => {
+describe('LloydGeorgeUploadDocumentsPage', () => {
```
joefong-nhs, line 83:
```suggestion
describe('LloydGeorgeUploadPage', () => {
```
maybe better to put the component name here, same as other tests

joefong-nhs (changes requested):
Gave a few suggestions, most are minor things but I think the one about window.clearInterval() call needs to be changed. Happy to have a chat anytime

app/src/helpers/requests/uploadDocuments.ts
```diff
@@ -180,5 +188,29 @@ const uploadDocuments = async ({
...
+    const body = {
+        files: [
+            {
+                reference: fileKey,
+                type: document.docType,
+                fields: { Uploading: uploadingState },
+            },
+        ],
+    };
+    await axios.post(updateUploadStateUrl, body, {
```
RioKnightleyNHS, line 209: return this here, also needs a try catch, even if the catch is empty it prevents the page from breaking

RioKnightleyNHS (changes requested):

app/src/pages/lloydGeorgeUploadPage/LloydGeorgeUploadPage.tsx
RioKnightleyNHS: agree

RioKnightleyNHS (approved):

joefong-nhs (approved):

PR #332, opened 2024-04-03
PRMDR-720 Add navigation improvements

RioKnightleyNHS:


app/src/components/layout/navLinks/NavLinks.tsx
```diff
@@ -8,25 +8,74 @@ import { useSessionContext } from '../../../providers/sessionProvider/SessionPro
...
+                style={{ color: 'white' }}
                 role="link"
                 className="clickable"
-                data-testid="logout-btn"
-                onClick={(e) => nav(e, routes.LOGOUT)}
+                onClick={(e) => nav(e, routes.START)}
             >
-                Log Out
+                Home
             </Header.NavItem>
```
joefong-nhs, line 35: Suggest to add `href` to each of the NavItems, so that they can be reached by tab and also enter key will fire the onClick event.
(We have a PR that adds keyboard access to buttons and links, but that PR can't cover the change of Nav menu in this branch)

app/src/components/layout/navLinks/NavLinks.tsx
```diff
@@ -8,25 +8,74 @@ import { useSessionContext } from '../../../providers/sessionProvider/SessionPro
...
+            {width <= 990 ? (
+                <>
+                    <Header.NavItem
+                        role="link"
+                        className="clickable"
+                        style={{ color: 'white' }}
```
joefong-nhs, line 41: When focused, this text will become white on yellow background which is poor in colour contrast.
Maybe better to make a css class to override style, so that we can config &:focus and possible other stuffs in one place

app/src/components/layout/navLinks/NavLinks.tsx
```diff
@@ -8,25 +8,74 @@ import { useSessionContext } from '../../../providers/sessionProvider/SessionPro
...
     const navigate = useNavigate();
     const [session] = useSessionContext();
-
+    const [width, setWidth] = useState<number>(window.innerWidth);
     const nav = (e: ReactEvent<HTMLAnchorElement, MouseEvent>, link: string) => {
         e.preventDefault();
         navigate(link);
     };
 
+    console.log(width);
```
joefong-nhs, line 17:
```suggestion
```

app/src/components/layout/navLinks/NavLinks.tsx
```diff
@@ -8,25 +8,74 @@ import { useSessionContext } from '../../../providers/sessionProvider/SessionPro
...
+            window.removeEventListener('resize', handleResizeWindow);
+        };
+    }, []);
+
     return session.isLoggedIn ? (
-        <Header.Nav>
-            <Header.NavItem role="link" className="clickable" onClick={(e) => nav(e, routes.HOME)}>
-                Home
-            </Header.NavItem>
+        <Header.Nav style={{ display: 'initial', color: 'white' }}>
```
joefong-nhs, line 27: Not directly related to this line, but within this Header.Nav component they got an embedded close button which only appears when width < 990 (see below pic)
image

As now we override the css and make this Nav to appear all the time, I would suggest to override the class "nhsuk-header navigation-close" to display:none, so that we don't have a close button that doesn't serve any purpose

joefong-nhs (approved):
checked in local that the nav menu works nicely :D

PR #334, opened 2024-04-04
PRMDR-758 - Keyboard Accessible Skip-Links

joefong-nhs:
( draft PR to trigger sonar cloud scan )

### DIFF
--- a/app/src/helpers/utils/manageFocus.ts
+++ b/app/src/helpers/utils/manageFocus.ts
@@ -17,3 +17,10 @@ export function focusElement(element: HTMLElement): void {
 
     element.focus();
 }
+
+export function focusLayoutDiv() {
+    const layoutDiv = document.querySelector('div#layout');
+    if (layoutDiv) {
+        (layoutDiv as HTMLDivElement).focus();
+    }
+}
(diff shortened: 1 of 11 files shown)


app/src/helpers/utils/manageFocus.ts
```diff
@@ -0,0 +1,26 @@
...
+            },
+            { once: true },
+        );
+    }
+
+    element.focus();
+}
+
+export function focusLayoutDiv() {
+    const layoutDiv = document.querySelector('div#layout');
```
RioKnightleyNHS, line 22: Would it be better to pass an input ref of the H1 instead ? Generally bad practise to use query Selector in react as document can be undefined

RioKnightleyNHS (changes requested):

app/src/helpers/utils/manageFocus.ts
joefong-nhs: Thanks for spotting this!
As it seems quite a big task to keep a global context for h1 ref and populate it on every page, for now I have addressed the problem like this:
 - use `document?.` to prevent any query causing an error
 - add a comment in relevant code to raise attention in case we use SSR in the future (which would break usages of `document` object)

RioKnightleyNHS (approved):

PR #335, opened 2024-04-05
Prmdr 762 - auth feature flag

NogaNHS:


lambdas/handlers/back_channel_logout_handler.py
```diff
@@ -22,8 +24,16 @@
...
+    feature_flag_service = FeatureFlagService()
+    auth_flag_name = FeatureFlags.USE_SMARTCARD_AUTH.value
+    use_smartcard_lambda_enabled_flag_object = (
+        feature_flag_service.get_feature_flags_by_flag(auth_flag_name)
+    )
+    if use_smartcard_lambda_enabled_flag_object[auth_flag_name]:
+        auth_ssm_prefix = "/auth/smartcard/"
+    else:
+        auth_ssm_prefix = "/auth/password/"
+    back_channel_logout_service = BackChannelLogoutService(auth_ssm_prefix)
```
joefong-nhs, line 36: Would it be possible to just initiate the FeatureFlagService() at oidc_service and load the flag there?  It seems a bit odd that we have to load a hard-coded string at the top level, and then pass that all along to a bottom-level

NogaNHS: I was thinking about it too, but it seems odd to have a configuration nested in the oidc service.

joefong-nhs: all right, thanks!

lambdas/handlers/login_redirect_handler.py
```diff
@@ -13,13 +15,31 @@
...
+        "OIDC_CALLBACK_URL",
+    ]
+)
 @handle_lambda_exceptions
 def lambda_handler(event, context):
     request_context.app_interaction = LoggingAppInteraction.LOGIN.value
     logger.info("Login Redirect handler triggered")
 
-    login_redirect_service = LoginRedirectService()
+    feature_flag_service = FeatureFlagService()
```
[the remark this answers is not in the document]
SRAlexander: Right, my recommendation is to remove this out to a DynamicConfigurationService to remove this duplication.
We discussed that it's declared in the handler and passed in as a service parameter which is against best practises but you had concerns that other dependent services also used this param. Have a review to see if the service parameter can be removed with the aid of this service, however, ensure that we only make a single call to the param store via the service, if we're making multiple calls for the same param, maintain the service creation parameter.

SRAlexander (changes requested):

lambdas/services/dynamic_configuration_service.py
```diff
@@ -0,0 +1,22 @@
...
+    def set_auth_ssm_prefix(self) -> None:
+        auth_flag_name = FeatureFlags.USE_SMARTCARD_AUTH.value
+        use_smartcard_lambda_enabled_flag_object = (
+            self.feature_flag_service.get_feature_flags_by_flag(auth_flag_name)
+        )
+        if use_smartcard_lambda_enabled_flag_object[auth_flag_name]:
+            request_context.auth_ssm_prefix = "/auth/smartcard/"
+        else:
+            request_context.auth_ssm_prefix = "/auth/password/"
+        logger.info("Setting auth ssm prefix to" + request_context.auth_ssm_prefix)
```
SRAlexander, line 22: Do we need a space before the " for a clean message structure?

SRAlexander (approved):

PR #338, opened 2024-04-09
PRMDR-795, PRMDR-796  Fix http status code at UpdateUploadState Lambda

joefong-nhs:
Fixed the http status code that we use in UpdateUploadState Lambda, according to the description in ticket PRMDR-741


SRAlexander (approved):
Could you check the UI and make sure there's no specific status code check taking place before merging @joefong-nhs.

If not, all good to merge.

PR #345, opened 2024-04-18
PRMDR 786 - update to title page

NogaNHS:


app/src/components/layout/pageTitle/PageTitle.tsx
```diff
@@ -0,0 +1,14 @@
...
+};
+
+function PageTitle({ pageTitle }: Props) {
+    const serviceName = 'Digital Lloyd George records';
+    useEffect(() => {
+        document.title = pageTitle + ' - ' + serviceName;
+    }, [pageTitle]);
+}
+
+export default PageTitle;
```
joefong-nhs, line 14: Could we be clear about whether this function is intended as a hook function or a react component?
Usually, a component has Capital case naming, has a return value that render some html, and is used in the return part of other components
A hook function has the naming convention "useXxx", may not have a return value, and is used before the return part of other components
For now, the usage of this function looks like a hook, but it is put in the component folder, and there is a mix of `PageTitle` and `pageTitle` where it is referred.

app/cypress/e2e/0-ndr-core-tests/gp_user_workflows/patient_search_and_verify_workflow.cy.js
```diff
@@ -63,6 +65,7 @@ describe('GP Workflow: Patient search and verify', () => {
                     'Check these patient details match the records or attachments you plan to use',
                 );
                 cy.get('#verify-submit').click();
+                cy.title().should('eq', 'Available records - Digital Lloyd George records');
```
joefong-nhs, line 68: This test seems to be failing. Looks like when the patient is inactive, the logic at PatientResultPage.tsx line 32:36 would redirect us to the old ARF upload page instead of LG view page.

PR #348, opened 2024-04-23
PRMDR-794 

NogaNHS:


app/src/styles/App.scss
```diff
@@ -134,7 +134,7 @@ $govuk-compatibility-govukelements: true;
     &_record-stage {
         &_patient-info {
             p {
-                font-size: 19px;
+                font-size: 1.2rem;
```
joefong-nhs, line 137: Would it be good to use the same rem value for 19 px between this line and LloydGeorgeFileInputStage.tsx? 
Here we have 1.2, while that file is using 1.188rem

joefong-nhs: sorry also line 385 of App.scss as well

joefong-nhs (approved):

PR #358, opened 2024-05-01
PRMDR - pilot fixes

NogaNHS:


lambdas/models/pds_models.py
```diff
@@ -66,10 +66,10 @@ class Patient(BaseModel):
 
     id: str
     birth_date: date
-    address: Optional[list[Address]] = []
+    address: Optional[list[Address]] = None
```
joefong-nhs, line 69: Agree to use None here as the default, but we may need to amend the method `get_current_home_address` at the same time, as the line `for entry in self.address:` assumes self.address to be an array.

lambdas/utils/lloyd_george_validator.py
```diff
@@ -140,31 +140,56 @@ def validate_filename_with_patient_details(
...
+    file_patient_first_name = patient_name_split[0]
+    file_patient_last_name = patient_name_split[-1]
+    is_file_first_name_in_patient_details = False
+    for patient_name in pds_patient_details.given_name:
+        if (
+            names_are_matching(file_patient_first_name, patient_name)
+            or not patient_name
+        ):
+            is_file_first_name_in_patient_details = True
+            break
```
joefong-nhs, line 165: seems that this part will set the value to True when pds_patient_details.given_name array contains a falsey value (e.g. 0, empty string, None), is that on purpose?

lambdas/models/pds_models.py
NogaNHS: Missed that one, thanks!

PR #366, opened 2024-05-16
PRMDR-881 - Remove personal information from file names of stitched LG record

joefong-nhs:
- For the stitched LG record pdf, use a random UUID as the file name
- Change the S3 path of stitched LG files from `nhs_number/filename` to `combined_files/filename`, so that URL will not contain patient NHS number anymore

### DIFF
--- a/lambdas/services/lloyd_george_stitch_service.py
+++ b/lambdas/services/lloyd_george_stitch_service.py
@@ -55,17 +55,15 @@ class LloydGeorgeStitchService:
             )
 
         try:
-            filename_for_stitched_file = self.make_filename_for_stitched_file(
-                lg_records
-            )
             stitched_lg_record = stitch_pdf(all_lg_parts, self.temp_folder)
+            filename_for_stitched_file = os.path.basename(stitched_lg_record)
             number_of_files = len(all_lg_parts)
             last_updated = self.get_most_recent_created_date(lg_records)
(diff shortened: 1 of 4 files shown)


lambdas/services/lloyd_george_stitch_service.py
```diff
@@ -55,23 +55,21 @@ def stitch_lloyd_george_record(self, nhs_number: str) -> str:
             )
 
         try:
-            filename_for_stitched_file = self.make_filename_for_stitched_file(
-                lg_records
-            )
             stitched_lg_record = stitch_pdf(all_lg_parts, self.temp_folder)
+            filename_for_stitched_file = os.path.basename(stitched_lg_record)
```
SRAlexander, line 59: Does this not still contain the patient's name?
I've got a couple of checks on this one....

- Do we reload the same file more than once or do we recreate the link every time? if We do reload the file more than once, should we hash the filename so we can generate it predictably?

SRAlexander (changes requested):

lambdas/services/lloyd_george_stitch_service.py
SRAlexander: Discussed, this will be an improvement down the line once we can monitor LG last upload state in code.

SRAlexander (approved):

PR #369, opened 2024-05-23
PRMDR-898 - Missing Unit tests on optional fields

joefong-nhs:


lambdas/models/pds_models.py
```diff
@@ -80,41 +83,57 @@ def get_security(self) -> Security:
...
-            givenName=self.get_current_usual_name().given,
-            familyName=self.get_current_usual_name().family,
+            givenName=given_name,
+            familyName=family_name,
             birthDate=self.birth_date,
             postalCode=(
-                self.get_current_home_address().postal_code
-                if self.is_unrestricted() and self.address
+                current_home_address.postal_code
+                if self.is_unrestricted() and current_home_address
```
NogaNHS, line 136: We're already checking `self.is_unrestricted()` in `self.get_current_home_address()`

lambdas/models/pds_models.py
```diff
@@ -15,13 +18,13 @@ class Address(BaseModel):
 
     use: str
     period: Optional[Period] = None
-    postal_code: Optional[str] = ""
+    postal_code: str = ""
```
NogaNHS, line 21: Postal code is an optional field

lambdas/models/pds_models.py
joefong-nhs: Thanks!  I've removed the duplicated check here

NogaNHS (approved):

PR #370, opened 2024-05-24
PRMDR:879 - Implement lambda layers into workflow

abbas-khan10:


Makefile
```diff
@@ -26,12 +40,14 @@ format:
 	./lambdas/venv/bin/ruff check lambdas/ --fix
 
 sort-requirements:
-	sort -o lambdas/requirements.txt lambdas/requirements.txt
-	sort -o lambdas/requirements-test.txt lambdas/requirements-test.txt
+	sort -o $(TEST_REQUIREMENTS) $(TEST_REQUIREMENTS)
+	sort -o $(CORE_REQUIREMENTS) $(CORE_REQUIREMENTS)
+	sort -o $(CORE_REQUIREMENTS) $(DATA_REQUIREMENTS)
```
joefong-nhs, line 45:
```suggestion
	sort -o $(DATA_REQUIREMENTS) $(DATA_REQUIREMENTS)
```

lambdas/scripts/lambda_layer_update.py
```diff
@@ -0,0 +1,63 @@
...
+    def update_lambda(self):
+        print(f"Updating {self.function_name_aws} with new layer ARNs...")
+        self.client.update_function_configuration(
+            FunctionName=self.function_name_aws,
+            Layers=list(self.updated_lambda_arns.values()),
+        )
+
+
+if __name__ == "__main__":
+    LambdaLayerUpdate.start(LambdaLayerUpdate())
```
joefong-nhs, line 62: this line looks a little bit unfamiliar to me, would it be any different if we just init this object and call .start() ? 
```
update = LambdaLayerUpdate()
update.start()
```

Makefile
abbas-khan10: Good spot thank you

lambdas/scripts/lambda_layer_update.py
abbas-khan10: Nope same logic just shorthand for instantiating and running method call in one line, I guess for a readability perspective we could rewrite this line so it's clearer

lambdas/requirements/layers/requirements_core_lambda_layer.txt
```diff
@@ -0,0 +1,24 @@
+PyJWT==2.8.0
```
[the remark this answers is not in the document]
abbas-khan10: Forgot about this cheers

lambdas/scripts/lambda_layer_update.py
```diff
@@ -0,0 +1,64 @@
...
+        print("")
+
+    def start(self):
+        self.extract_default_layer_arns()
+        self.extract_updated_layer_arns()
+        self.update_lambda()
+
+    def extract_default_layer_arns(self):
+        response = self.client.get_function(FunctionName=self.function_name_aws)
+        layer_arns_responses = response["Configuration"]["Layers"]
```
NogaNHS, line 33:
```suggestion
        layer_responses = response["Configuration"]["Layers"]
```
As this is a list of layers not arns

lambdas/scripts/lambda_layer_update.py
```diff
@@ -0,0 +1,64 @@
...
+
+        for layer_arn_response in layer_arns_responses:
+            layer_arn = layer_arn_response["Arn"]
+            match = re.search(r"layer:([^:]+):", layer_arn)
+            if match:
+                layer_name = match.group(1)
+
```
NogaNHS, line 40: What are we checking here?

.github/workflows/base-lambda-layer-reusable-publish-all.yml
```diff
@@ -0,0 +1,63 @@
+name: 'ZBASE Lambda Layers: Deploy all required Lambda layers'
```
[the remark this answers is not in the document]
abbas-khan10: This is helpful as the wokflows are alphabetically ordered, and since this base modules are used by others rather than being the main wokflows to be triggered, they will be ordered at the bottom as they have a Z prefix

lambdas/scripts/lambda_layer_update.py
abbas-khan10: We're checking for a match between "lambda:" and the next ":" to extract the layer name from the layer arn. I've updated the "match" variable as it's not entirely clear

NogaNHS (approved):

lambdas/scripts/lambda_layer_update.py
```diff
@@ -0,0 +1,64 @@
...
+        layer_responses = response["Configuration"]["Layers"]
+
+        for layer_response in layer_responses:
+            layer_arn = layer_response["Arn"]
+            layer_name_match = re.search(r"layer:([^:]+):", layer_arn)
+            if layer_name_match:
+                layer_name = layer_name_match.group(1)
+
+                if layer_name in ["AWS-AppConfig-Extension", "LambdaInsightsExtension"]:
+                    self.updated_lambda_arns.update({layer_name: layer_arn})
```
NogaNHS, line 42: Maybe we should remove the default layers from terraform and hard code them here?

PR #374, opened 2024-06-07
PRMDR-914 new IAM service for pre-sign url

NogaNHS:


lambdas/services/create_document_reference_service.py
```diff
@@ -30,7 +30,8 @@
 
 class CreateDocumentReferenceService:
     def __init__(self):
-        self.s3_service = S3Service()
+        create_document_aws_role_arn = os.getenv("PRE_SIGN_ASSUME_ROLE")
```
joefong-nhs, line 33: probably better to add this new env var to the handler's decorator, so that in case this is missing we are notified

lambdas/services/create_document_reference_service.py
```diff
@@ -30,7 +30,8 @@
...
+        self.s3_service = S3Service(custom_aws_role=create_document_aws_role_arn)
```
joefong-nhs, line 34: Just an idea, would it be good to change this `custom_aws_role` as a new argument of presign url method e.g. create_upload_presigned_url, rather than passing it during service initialisation?

The reasons that I suggest this:
 - If another lambda wants to use presign url, the method signature for e.g `create_upload_presigned_url` would clearly indicate that a `custom_aws_role` is necessary for that method to work
 - S3Service can be initialised in the same way `S3Service()` throughout the codebase, which would be better as that service is singleton

lambdas/services/base/s3_service.py
```diff
@@ -12,41 +13,50 @@
...
                 signature_version="s3v4",
             )
-            self.client = boto3.client("s3", config=config)
             self.presigned_url_expiry = 1800
+            self.client = boto3.client("s3", config=config)
             self.initialised = True
+            self.custom_client = None
+            if custom_aws_role:
+                iam_service = IAMService()
+                print(iam_service)
```
joefong-nhs, line 35: probably don't need this print anymore :)

lambdas/services/lloyd_george_stitch_service.py
```diff
@@ -34,7 +34,8 @@ def __init__(self):
             "STITCHED_FILE_LIFECYCLE_POLICY_TAG", "autodelete"
         )
 
-        self.s3_service = S3Service()
+        create_document_aws_role_arn = os.getenv("PRE_SIGN_ASSUME_ROLE")
+        self.s3_service = S3Service(custom_aws_role=create_document_aws_role_arn)
```
joefong-nhs, line 38:
```suggestion
       presign_url_aws_role_arn = os.getenv("PRE_SIGN_ASSUME_ROLE")
        self.s3_service = S3Service(custom_aws_role=presign_url_aws_role_arn)
```
suggest to rename as here the role isn't for creating document

lambdas/services/base/iam_service.py
```diff
@@ -0,0 +1,32 @@
+import boto3
+from botocore.exceptions import ClientError
+from utils.audit_logging_setup import LoggingService
+
+logger = LoggingService(__name__)
+
+
+class IAMService:
+    def __init__(self):
```
abbas-khan10, line 9: I wonder if we should configure this to be a singleton also? For now it's only the s3 service using it, however because the s3 service is a singleton it will spawn multiple instances of this iam service when we only need one

lambdas/services/base/s3_service.py
```diff
@@ -12,41 +13,49 @@
 class S3Service:
     _instance = None
 
-    def __new__(cls):
+    def __new__(cls, *args, **kwargs):
```
abbas-khan10, line 16: Any reason for the args/kwargs addition?

lambdas/services/base/s3_service.py
```diff
@@ -12,41 +13,49 @@
...
+            self.client = boto3.client("s3", config=config)
             self.initialised = True
+            self.custom_client = None
+            if custom_aws_role:
+                iam_service = IAMService()
+                self.custom_client = iam_service.assume_role(
+                    custom_aws_role, "s3", config=config
+                )
 
     # S3 Location should be a minimum of a s3_object_key but can also be a directory location in the form of
```
abbas-khan10, line 39: Wonder if we still need this comment? I know it's not part of your work but seems a bit unnecessary to me?

lambdas/services/base/s3_service.py
NogaNHS: `__new__` is run before `__init__` and all the args are passed into both. So I need to be able to pass the custom role into `__init__`

NogaNHS: When __new__() returns an instance of cls it automatically invokes the [__init__()] method of that instance with the arguments passed to it.(https://www.codecademy.com/resources/docs/python/dunder-methods/new)

lambdas/services/base/s3_service.py
NogaNHS: I think we do. It's not clear without it.

PR #376, opened 2024-06-10
PRMP-198_199 - Lambdas: DataCollectionLambda and StatisticalReportLambda

joefong-nhs:


.github/workflows/base-lambdas-reusable-deploy-all.yml
```diff
@@ -316,4 +316,32 @@ jobs:
...
+    name: Deploy data collection lambda
+    uses: ./.github/workflows/base-lambdas-reusable-deploy.yml
+    with:
+      environment: ${{ inputs.environment}}
+      python_version: ${{ inputs.python_version }}
+      build_branch: ${{ inputs.build_branch}}
+      sandbox: ${{ inputs.sandbox }}
+      lambda_handler_name: data_collection_handler
+      lambda_aws_name: DataCollectionLambda
+      lambda_layer_names: "core_lambda_layer,data_lambda_layer"
```
abbas-khan10, line 331: Do both these lambdas require polars?

lambdas/models/cloudwatch_logs_query.py
```diff
@@ -0,0 +1,53 @@
+from dataclasses import dataclass
```
abbas-khan10, line 1: I wonder if this belongs in the utils folder alongside the dynamo filters, as this is not really a models file?

.github/workflows/base-lambdas-reusable-deploy-all.yml
joefong-nhs: Yes, I have used polars in the DataCollection lambda for a few metrics that are a bit tricky and need some calculation, for example, "average number of records per patient per each ODS code"

lambdas/models/cloudwatch_logs_query.py
joefong-nhs: Thanks! Agree that this makes more sense to be in the utils folder. I will move them now.

lambdas/services/data_collection_service.py
```diff
@@ -0,0 +1,324 @@
...
+
+        record_store_data = self.get_record_store_data(
+            dynamodb_scan_result, s3_list_objects_result
+        )
+        organisation_data = self.get_organisation_data(dynamodb_scan_result)
+        application_data = self.get_application_data()
+
+        return record_store_data + organisation_data + application_data
+
+    def write_to_local_dynamodb_table(self, all_statistic_data: list[StatisticData]):
```
abbas-khan10, line 73: Why have we included 'local' here?

lambdas/services/data_collection_service.py
```diff
@@ -0,0 +1,324 @@
...
+        self.dynamodb_service.batch_writing(
+            table_name=self.output_table_name, item_list=item_list
+        )
+
+        logger.info("Finish writing all data to dynamodb table")
+
+    def scan_dynamodb_tables(self) -> list[dict]:
+        all_results = []
+
+        project_expression = "CurrentGpOds,NhsNumber,FileLocation,ContentType"
```
abbas-khan10, line 88: Should have have these hardcoded here? Perhaps a string interpolation of our enum could be a better option?

lambdas/services/data_collection_service.py
```diff
@@ -0,0 +1,324 @@
...
+            ods_code = item["CurrentGpOds"]
+            file_type = item["ContentType"]
+            file_types_grouped_by_ods_code[ods_code].add(file_type)
+
+        return [
+            {"ods_code": ods_code, "number_of_document_types": len(file_types)}
+            for ods_code, file_types in file_types_grouped_by_ods_code.items()
+        ]
+
+    def get_average_number_of_file_per_patient(
```
abbas-khan10, line 295: *files

lambdas/services/data_collection_service.py
joefong-nhs: sorry, this was a legacy name when I built this with a locally hosted dynamodb
renamed to `write_to_dynamodb_table` now

lambdas/services/data_collection_service.py
joefong-nhs: Thanks, totally agree that we should use our enum instead.
I have replaced this line and all other hardcoded field names of DocumentReference in this class with values from enum

lambdas/services/data_collection_service.py
joefong-nhs: Nice catch! :>  Fixed just now

abbas-khan10 (approved):

PR #378, opened 2024-06-11
PRMP-390 LG Download number of files fix

NogaNHS:


app/cypress/e2e/0-ndr-core-tests/gp_user_workflows/view_lloyd_george_not_bsol_workflow.cy.js
```diff
@@ -261,9 +261,7 @@ describe('GP Workflow: View Lloyd George record', () => {
 
                     // Assert contents of page when downloading
                     cy.contains('Downloading documents').should('be.visible');
-                    cy.getByTestId(
-                        'download-file-header-' + viewLloydGeorgePayload.number_of_files + '-files',
-                    ).should('be.visible');
+                    cy.getByTestId('download-file-header-0-files').should('be.visible');
```
joefong-nhs, line 264: Is it correct to expect test id "download-file-header-0-files"? 
From the code that generates this test id, it seems that `download-file-header-0-files` indicates the text showing "Preparing download for 0 files", which is the issue that we should probably fix.
When I actually run this cypress test, the screen shows "download for 0 files" as well
image

NogaNHS: This is because the non-bsol route doesn't call the get documents list lambda. So we don't have the search results at any point, which is what we use for the documents number.
Before the UI was using the number that is returns from the stitch lambda about the number of files.
So to fix this we would need to pass in again the number of file from the stitch lambda. 
The reason I change this to 0 was that the only time I found the file number is being used in the non-Bsol workflow was in the test-id. But if we also use it for this screen I'll change it back.

NogaNHS: Done

app/src/components/blocks/_lloydGeorge/lloydGeorgeSelectDownloadStage/LloydGeorgeSelectDownloadStage.tsx
```diff
@@ -119,9 +117,15 @@ function LloydGeorgeSelectDownloadStage({
             mounted.current = true;
             void onPageLoad();
         }
+        numberOfFilesForDownload.current = selectedDocuments?.length
+            ? selectedDocuments.length
+            : searchResults.length
+            ? searchResults.length
+            : numberOfFilesForDownload.current;
```
joefong-nhs: Putting this calculation in useEffect might give unexpected result in some case, 
for example, when I have 5 files and selected only 3 files, then click the "download-all" button (I tested this on a sandbox), it seems to display "preparing 3 files", which isn't correct (as I downloaded all 5 files).

Maybe we better not use a ref to store this value. Just use a normal variable, and do the calculation out of useEffect, for example:
```
   // replace line 81 `let numberOfFilesForDownload = useRef(numberOfFiles)` with below:
    const numberOfFilesForDownload =
        selectedDocuments?.length || searchResults.length || numberOfFiles;
```

demo file number issue

joefong-nhs (approved):

PR #384, opened 2024-06-19
PRMP-535 - Changes to createDocRef lambda

joefong-nhs:


lambdas/services/create_document_reference_service.py
```diff
@@ -287,7 +292,16 @@ def remove_records_of_failed_lloyd_george_upload(
...
         logger.info("Previous failed records are deleted.")
+
+    def fetch_incomplete_arf_upload_records(
+        self, nhs_number
+    ) -> list[DocumentReference]:
+        return self.document_service.fetch_available_document_references_by_type(
+            nhs_number=nhs_number,
+            doc_type=SupportedDocumentTypes.ARF,
+            query_filter=UploadIncomplete,
+        )
```
NogaNHS, line 307: Can we change this to `fetch_incomplete_records` and add a parameter for doc_type, and then call this fun in `check_existing_lloyd_george_records`?

lambdas/services/create_document_reference_service.py
```diff
@@ -141,21 +149,17 @@ def prepare_doc_object(
 
         logger.info(PROVIDED_DOCUMENT_SUPPORTED_MESSAGE)
 
-        if validated_doc.docType == SupportedDocumentTypes.LG.value:
+        if validated_doc.docType in [
+            SupportedDocumentTypes.LG.value,
+            SupportedDocumentTypes.ARF.value,
+        ]:
```
NogaNHS, line 155: Is this check necessary? if it is not one of those it would have thrown a ValidationError before.

lambdas/services/create_document_reference_service.py
```diff
@@ -103,6 +102,15 @@ def create_document_reference_request(
                 )
```
NogaNHS, line 102: Can't comment those lines, but could we move the check on line 98 for existing LG records to line 65?

lambdas/services/create_document_reference_service.py
```diff
@@ -103,6 +102,15 @@ def create_document_reference_request(
...
+                )
+
+                self.stop_if_upload_is_in_process(incomplete_arf_upload_records)
+                self.remove_records_of_failed_upload(
+                    self.arf_dynamo_table, incomplete_arf_upload_records
+                )
+
                 self.create_reference_in_dynamodb(
                     self.arf_dynamo_table, arf_documents_dict_format
                 )
```
NogaNHS, line 116: If it's ok, could you move this to a separated function? this `create_document_reference_request` is getting too big.

lambdas/services/create_document_reference_service.py
NogaNHS, line 1: On line 67 we call `getting_patient_info_from_pds(nhs_number)` this can raise a Validation error from the pds, and need to be caught

lambdas/services/create_document_reference_service.py
joefong-nhs: Hi, for this one, I think the check on line 98 might be better to be after LG filename checks on line 97,
The reason is that:
1. the check will make API call to dynamodb table, so better to run all locally computed validation first
2. at the end of the check in line 98, it will remove incomplete upload records

So I am leaning towards keeping it on line 98, but I will rename the function at line 98 `check_existing_lloyd_george_records` --> `check_existing_lloyd_george_records_and_remove_failed_upload` so that the function name better reflects what it does.

lambdas/services/create_document_reference_service.py
joefong-nhs: Thanks! Yes we don't need this check anymore. Will take this away
 Sorry I just combined the previous if-else branches and didn't notice this 😓

lambdas/services/create_document_reference_service.py
joefong-nhs: Thanks, got this in a separate function now :)

lambdas/services/create_document_reference_service.py
joefong-nhs: Thank for spoting this, it is an important one
I have made the below changes:
1. In the method getting_patient_info_from_pds, catch and log pydantic Validation error and then raise LGInvalidFilesException
2. Move this call within the try-catch block of create doc ref, so that the error will be caught and converted to 400 response
Also I've added a few unit tests to verify the changes

NogaNHS (approved):

PR #391, opened 2024-07-05
PRMP 566 - GenerateDocumentManifest lambda

steph-torres-nhs:


lambdas/handlers/generate_document_manifest_handler.py
```diff
@@ -0,0 +1,79 @@
...
+def prepare_zip_trace_data(new_zip_trace: dict) -> dict:
+
+    for key, nested_object in new_zip_trace.items():
+        value = list(nested_object.values())[0]
+        if isinstance(value, dict):
+            prepare_zip_trace_data(value)
+
+        new_zip_trace[key] = list(nested_object.values())[0]
+
+    return new_zip_trace
```
joefong-nhs, line 79: for this one, might be better to use the default deserializer in boto3 rather than writing our own
e.g. 

```suggestion
def prepare_zip_trace_data(new_zip_trace: dict) -> dict:
    deserialize = TypeDeserializer().deserialize

    parsed_dynamodb_items = {
        key: deserialize(dynamodb_value)
        for key, dynamodb_value in new_zip_trace.items()
    }

    return parsed_dynamodb_items
```

TypeDeserializer is a class from boto3
```
from boto3.dynamodb.types import TypeDeserializer
```

lambdas/handlers/generate_document_manifest_handler.py
```diff
@@ -0,0 +1,79 @@
...
+
+    for record in dynamo_records:
+        new_zip_trace = record.get("dynamodb", {}).get("NewImage")
+        event_name = record.get("eventName")
+        if (
+            not new_zip_trace
+            or event_name != "INSERT"
+            or not isinstance(new_zip_trace, dict)
+        ):
+            return ApiGatewayResponse(400, "", "GET").create_api_gateway_response()
```
joefong-nhs, line 45: Would it be good to put some error logging before the return statement of line 35 and line 45?   
Because probably the 400 errors won't be received by anyone, so putting some error log here would help us to detect when any issue occur

lambdas/services/base/dynamo_service.py
```diff
@@ -78,23 +78,15 @@ def query_with_requested_fields(
             logger.error(str(e), {"Result": f"Unable to query table: {table_name}"})
             raise e
 
-    def query_all_fields(self, table_name: str, key_condition_expression):
+    def query_all_fields(self, table_name: str, search_key, search_condition):
```
abbas-khan10, line 81: Can we add types here?

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -0,0 +1,120 @@
...
+    def update_processing_status(self):
+        self.zip_trace_object.job_status = ZipTraceStatus.PROCESSING
+        self.update_dynamo_with_fields({"job_status"})
+
+    def update_failed_status(self):
+        self.zip_trace_object.job_status = ZipTraceStatus.FAILED
+        self.update_dynamo_with_fields({"job_status"})
+
+    def remove_temp_files(self):
+        # Removes the parent of each removed directory until the parent does not exist or the parent is not empty
```
abbas-khan10, line 118: Is this comment necessary? I would maybe rename this function to "cleanup_temp_files" or something similar to describe what is happening here

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -0,0 +1,120 @@
...
+            msg = f"{file_key} may reference missing file in s3 bucket: {file_bucket}"
+            logger.error(
+                f"{LambdaError.ZipServiceClientError.to_str()} {msg + str(e)}",
+                {"Result": "Failed to create document manifest"},
+            )
+            raise DocumentManifestServiceException(
+                status_code=500, error=LambdaError.ZipServiceClientError
+            )
+
+    def get_file_bucket_and_key(self, file_location):
```
abbas-khan10, line 61: type missing in parameters

abbas-khan10 (approved):

PR #393, opened 2024-07-08
PRMP-568 - Modify the Front End to handle new poll based download logic

joefong-nhs:


app/cypress/e2e/0-ndr-core-tests/gp_user_workflows/download_lloyd_george_workflow.cy.js
```diff
@@ -59,10 +59,17 @@ describe('GP Workflow: View Lloyd George record', () => {
...
+                statusCode: 200,
+                body: { jobId: 'test-jobId' },
+            });
+        }).as('documentManifestPost');
+
         cy.intercept('GET', '/DocumentManifest*', (req) => {
             req.reply({
                 statusCode: 200,
-                body: baseUrl + '/browserconfig.xml', // uses public served file in place of a ZIP file
+                body: { jobStatus: 'Completed', url: baseUrl + '/browserconfig.xml' }, // uses public served file in place of a ZIP file
```
NogaNHS, line 72: Could you add a test case for if the jobStatus is still pending and the url is empty?

joefong-nhs: Thanks, I have added the pending to the base test case. The intercept will now respond with "Processing" and an empty url for a few times before it responds with "Complete"

app/src/helpers/requests/getPresignedUrlForZip.ts
```diff
@@ -11,30 +17,85 @@ type Args = {
...
-type GetPresignedUrl = {
-    data: string;
+type GetRequestArgs = {
+    jobId: string;
+    baseUrl: string;
+    baseHeaders: AuthHeaders;
+};
+
+const ThreePendingErrorMessage =
+    'Document Manifest api call responded with "Pending" for 3 attempts.';
```
steph-torres-nhs, line 27: Maybe a different message, "Failed to initiate download"?

app/cypress/e2e/0-ndr-core-tests/gp_user_workflows/download_lloyd_george_workflow.cy.js
joefong-nhs: Also add a test case for getting "Pending" 3 times and display an error page

app/src/helpers/requests/getPresignedUrlForZip.ts
joefong-nhs: thanks! fixed this just now

NogaNHS (approved):

PR #397, opened 2024-07-19
PRMP-306 - ARF upload with the same file names is not handled correctly

joefong-nhs:


app/src/helpers/requests/uploadDocument.test.ts
```diff
@@ -37,6 +43,27 @@ describe('[POST] updateDocumentState', () => {
     });
 });
 
+describe('uploadDocumentToS3', () => {
```
joefong-nhs, line 46: note: I added this test because sonarcloud marked this function as untested and that affected the coverage %.

NogaNHS (approved):

PR #400, opened 2024-08-01
Prmp 637 - Screenreader interactions with toggle select all button and checkboxes

steph-torres-nhs:


app/src/components/blocks/_lloydGeorge/lloydGeorgeSelectSearchResults/LloydGeorgeSelectSearchResults.tsx
```diff
@@ -96,12 +123,17 @@ const AvailableFilesTable = ({
...
+                                            aria-checked={selectedDocuments.includes(result.ID)}
                                         >
-                                            <span className="nhsuk-u-visually-hidden">
+                                            <label
+                                                htmlFor={result.fileName}
+                                                className="nhsuk-u-visually-hidden"
+                                            >
                                                 {result.fileName}
-                                            </span>
+                                            </label>
```
joefong-nhs, line 136: If we look into the html this renders, the original Checkboxes.Box already create a <label> tag for us. Adding a <label> here will create a nested tag with an incorrect for= attribute, which probably affect screenreader functionality.
I would suggest to revert to previous version
image

app/src/components/blocks/_lloydGeorge/lloydGeorgeSelectSearchResults/LloydGeorgeSelectSearchResults.tsx
```diff
@@ -51,6 +51,23 @@ const AvailableFilesTable = ({
...
+    };
+
+    const handleToggleButtonStatusChange = () => {
+        if (selectedDocuments.length === searchResults.length) {
+            return 'All files are selected';
+        } else if (selectedDocuments.length === 0) {
+            return 'All files are deselected';
+        }
+    };
+
```
joefong-nhs, line 70: I would suggest renaming these two functions to something like "getAriaDescription", "getAnnouncementMessage". 
Usually in react we call a function `handleXXX` if it handles an event and triggers an action that affects other things, e.g. make an API call or update a state. Here these two function only provide a different string depending on situation, so probably better not to call them `handleXXX`

app/src/components/blocks/_lloydGeorge/lloydGeorgeSelectSearchResults/LloydGeorgeSelectSearchResults.test.tsx
```diff
@@ -158,7 +158,9 @@ describe('LloydGeorgeSelectSearchResults', () => {
                     expect(checkbox).toBeChecked();
                 });
 
-                expect(toggleSelectAllBtn.textContent).toBe('Deselect all files');
+                expect(toggleSelectAllBtn.textContent).toBe(
+                    'Deselect all filesAll files are selected',
+                );
```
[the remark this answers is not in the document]
steph-torres-nhs: Done

app/src/components/blocks/_lloydGeorge/lloydGeorgeSelectSearchResults/LloydGeorgeSelectSearchResults.tsx
steph-torres-nhs: Done

app/src/components/blocks/_lloydGeorge/lloydGeorgeSelectSearchResults/LloydGeorgeSelectSearchResults.tsx
steph-torres-nhs: Done

joefong-nhs (approved):

PR #413, opened 2024-08-20
PRMP - 763 Update patient ods using batchupdate script

steph-torres-nhs:


lambdas/services/batch_update_ods_code.py
```diff
@@ -196,31 +206,5 @@ def build_progress_dict(self, dynamodb_records: list[dict]) -> dict:
...
-    setup_logging_for_local_script()
-
-    parser = argparse.ArgumentParser(
-        prog="batch_update_ods_code.py",
-        description="A utility script to update the ODS Codes for all patients in a dynamoDB doc reference table",
-    )
-    parser.add_argument("table_name", type=str, help="The name of dynamodb table")
-    args = parser.parse_args()
-
-    BatchUpdate(table_name=args.table_name).main()
```
abbas-khan10, line 226: Why have we removed this main? And have we checked for usages of this script elsewhere if it used in the actions anywhere, it may break the process if we rely on the `batch_update.main()` it will work within a python environment, but if this script is called from anywhere else it will break. Is that the intended change?

lambdas/handlers/update_patient_ods_handler.py
```diff
@@ -0,0 +1,34 @@
...
+    names=[
+        "APPCONFIG_APPLICATION",
+        "APPCONFIG_CONFIGURATION",
+        "APPCONFIG_ENVIRONMENT",
+        "LLOYD_GEORGE_DYNAMODB_NAME",
+    ]
+)
+@override_error_check
+def lambda_handler(event, context):
+
```
NogaNHS, line 24: Remove space

lambdas/handlers/update_patient_ods_handler.py
```diff
@@ -0,0 +1,34 @@
...
+logger = LoggingService(__name__)
+
+
+@set_request_context_for_logging
+@ensure_environment_variables(
+    names=[
+        "APPCONFIG_APPLICATION",
+        "APPCONFIG_CONFIGURATION",
+        "APPCONFIG_ENVIRONMENT",
```
NogaNHS, line 18: needed?

lambdas/services/batch_update_ods_code.py
```diff
@@ -99,28 +97,40 @@ def update_patient_ods(self, nhs_number: str):
...
         pds_response = self.pds_service.pds_request(
             nhs_number=nhs_number, retry_on_expired=True
         )
         pds_response.raise_for_status()
 
-        patient = Patient.model_validate(pds_response.json())
-        return patient.get_active_ods_code_for_gp()
+        pds_response_json = pds_response.json()
+
+        patient = Patient.model_validate(pds_response_json)
```
NogaNHS, line 117: What if it's not valid? are we catching the error anywhere?

lambdas/services/batch_update_ods_code.py
```diff
@@ -99,28 +97,40 @@ def update_patient_ods(self, nhs_number: str):
...
+        ods_code = patient.get_active_ods_code_for_gp()
+
+        deceased = bool(pds_response_json.get("deceasedDateTime"))
+        if deceased:
+            return "DECE"
+
+        if not ods_code:
+            return "SUSP"
+
+        return ods_code
```
NogaNHS, line 128: I think this should be part of the PDS Model

lambdas/handlers/update_patient_ods_handler.py
AndyFlintNHS: @NogaNHS can you provide context for this please? Why does this stand out to you?

NogaNHS: We use app config for feature flags, and this lambda doesn't use any. (Maybe it should?)

lambdas/handlers/update_patient_ods_handler.py
```diff
@@ -0,0 +1,33 @@
...
+        "LLOYD_GEORGE_DYNAMODB_NAME",
```
NogaNHS, line 19: As we use PDS on this lambda, we should use the `PDS_FHIR_IS_STUBBED` env var

lambdas/services/batch_update_ods_code.py
```diff
@@ -35,28 +39,26 @@ def __init__(
         self.dynamo_service = DynamoDBService()
```
NogaNHS, line 39: Change pds_service to use `get_pds_service()`

NogaNHS (approved):

PR #416, opened 2024-08-27
[PRMP-839] Fixed suspended patient is active logic & added unit test

AndyFlintNHS:
Fixed bug where any truthy value for the ODS code would cause a patient to be flagged as active. This caused a problem where PCSE couldn't view suspended patients as their 'ODS code' was returning as `SUSP`, which is considered truthy and was incorrectly flagging them as active. 

Added a unit test with an example expected suspended patient response from PDS

### DIFF
--- a/lambdas/tests/unit/services/test_search_patient_details_service.py
+++ b/lambdas/tests/unit/services/test_search_patient_details_service.py
@@ -9,6 +9,11 @@ from utils.exceptions import (
 )
 from utils.lambda_exceptions import SearchPatientException
 
+USER_VALID_ODS_CODE = "ABC123"
+PATIENT_VALID_ODS_CODE = "XYZ789"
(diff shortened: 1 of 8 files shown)


lambdas/tests/unit/services/test_search_patient_details_service.py
```diff
@@ -9,6 +9,11 @@
 )
 from utils.lambda_exceptions import SearchPatientException
 
+USER_VALID_ODS_CODE = "ABC123"
+PATIENT_VALID_ODS_CODE = "XYZ789"
+INVALID_ODS_CODE = "SOME_INVALID_ODS_CODE"
```
NogaNHS, line 14: Not used

NogaNHS (approved):

bethany-kish-nhs (approved):

PR #420, opened 2024-09-10
PRMP - 895 - Accept deceased patients

NogaNHS:


lambdas/tests/unit/services/test_bulk_upload_service.py
```diff
@@ -33,6 +33,7 @@
 from tests.unit.helpers.data.pds.pds_patient_response import (
     PDS_PATIENT,
     PDS_PATIENT_DECEASED,
```
chrisbloe-nhse, line 35: Should the now be `PDS_PATIENT_DECEASED_FORMAL`?

lambdas/services/bulk_upload_service.py
```diff
@@ -115,25 +113,25 @@ def handle_sqs_message(self, message: dict):
...
                 raise LGInvalidFilesException("Patient not registered at your practice")
             patient_death_notification_status = (
                 pds_patient_details.get_death_notification_status()
             )
-            if is_formally_deceased(patient_death_notification_status):
-                raise PatientDeceasedException(
+            if patient_death_notification_status:
+                accepted_reason = (
                     f"Patient is deceased - {patient_death_notification_status.name}"
                 )
```
MohammadIqbalAD-NHS, line 130: If both `is_name_validation_based_on_historic_name` and `patient_death_notification_status` are true, the `accepted_reason` value assigned on Line 121 would be overwritten by Line 128. Would this pose a problem with what is saved to dynamo?

NogaNHS: Not sure what should take priority, let me check

MohammadIqbalAD-NHS (approved):

PR #421, opened 2024-09-10
prmdr786 - lambda zip deployment

RioKnightleyNHS:


lambdas/services/base/dynamo_service.py
```diff
@@ -118,22 +118,23 @@ def update_item(
...
+            "ExpressionAttributeValues": generated_expression_attribute_values,
         }
 
         if condition_expression:
             update_item_args["ConditionExpression"] = condition_expression
-            if expression_attribute_values:
-                expression_attribute_values.update(expression_attribute_values)
-                update_item_args["ExpressionAttributeValues"] = (
-                    expression_attribute_values
-                )
```
abbas-khan10, line 136: What was the purpose of this change?

PR #425, opened 2024-09-13
PRMP-754 - Bulk Upload: General Exception

MohammadIqbalAD-NHS:


steph-torres-nhs:
Morning, 
All looks good, is it worth adding testing/is there a way to see that the new ApiGatewayResponse with LambdaError is returned on unhandled errors?

steph-torres-nhs (approved):

PR #427, opened 2024-09-16
PRMP-885 Allow unnecessary spaces at the end of patient search input

steph-torres-nhs:


abid-nhs:
Hey Steph, 
I just realised that the ticket mentions:
"_Trim any spaces!  Make it work with any white space.
Validate any spaces work_"

The code doesn't currently fix the issue if there are 2 or more spaces after the nhs number which seems to be part of the scope of the ticket based on the above? Maybe worth discussing with product?

abid-nhs (approved):

bethany-kish-nhs (approved):

PR #437, opened 2024-09-26
PRMP-753 - Moving metadata CSV file to dated folder

AndyFlintNHS:


lambdas/services/bulk_upload_metadata_service.py
```diff
@@ -101,5 +105,21 @@ def send_metadata_to_fifo_sqs(
...
+        current_datetime = datetime.now().strftime("%Y-%m-%d_%H-%M")
+
+        self.s3_service.copy_across_bucket(
+            self.staging_bucket_name,
+            metadata_filename,
+            self.staging_bucket_name,
+            f"metadata/{current_datetime}.csv",
+        )
+
+        # delete the original once we've copied it to the dated folder
```
steph-torres-nhs, line 120: We can remove the comment if you like

lambdas/services/bulk_upload_metadata_service.py
```diff
@@ -101,5 +105,21 @@ def send_metadata_to_fifo_sqs(
                 group_id=sqs_group_id,
             )
 
+    def copy_metadata_to_dated_folder(self, metadata_filename: str):
+        logger.info("Copying metadata CSV to dated folder")
+
+        current_datetime = datetime.now().strftime("%Y-%m-%d_%H-%M")
+
+        self.s3_service.copy_across_bucket(
```
steph-torres-nhs, line 113: Might be worth copying the file from temp storage incase it is overwritten at any time, although I don't think this is currently possible, also not sure how we would handle the delete in this case

AndyFlintNHS: We've spoken about this - it's a good point and a consideration for a future change but not needed at the moment

steph-torres-nhs (approved):

PR #443, opened 2024-10-03
PRMP 988 - Remove error after deletion

RioKnightleyNHS:


app/src/pages/documentSearchResultsPage/DocumentSearchResultsPage.tsx
```diff
@@ -34,8 +34,13 @@ function DocumentSearchResultsPage() {
     const baseUrl = useBaseAPIUrl();
     const baseHeaders = useBaseAPIHeaders();
     const config = useConfig();
-
     const mounted = useRef(false);
+
+    const resetDocState = () => {
+        setSubmissionState(SUBMISSION_STATE.INITIAL);
+        setDownloadState(SUBMISSION_STATE.INITIAL);
```
NogaNHS, line 41: Do we need to set SearchResults to an empty list here?
Also should it be initial state? not failed?

app/cypress/e2e/0-ndr-core-tests/gp_user_workflows/view_lloyd_george_is_bsol_workflow.cy.js
```diff
@@ -216,15 +216,9 @@ describe('GP Workflow: View Lloyd George record', () => {
...
-                    { statusCode: 200, body: viewLloydGeorgePayload },
-                    { statusCode: 404 },
-                ];
-
-                cy.intercept('GET', '/LloydGeorgeStitch*', (req) => {
-                    req.reply(replies[request]);
-                    request = request + 1;
+                cy.intercept('GET', '/LloydGeorgeStitch*', {
+                    statusCode: 200,
+                    body: viewLloydGeorgePayload,
```
NogaNHS, line 221: Can we add a test where the first call to the stitch return 200 and then delete the record and check there is another call to the stitch lambda?

NogaNHS (approved):

PR #446, opened 2024-10-07
PRMP-844 refactoring column rename script

AndyFlintNHS:
### Discussion needed as to whether we'll merge this into main or not!

1. This PR is just to QA what I've done before we run it on any higher environments than the sandbox
2. I've branched this script off of PRMP-764 - this script was not merged into main so the git history makes it look like this is an entirely new file. You're best off comparing the changes between the PRMP-764 commit and the PRMP-844-script commit.

### DIFF
--- a/lambdas/scripts/bulk_upload_column_change.py
+++ b/lambdas/scripts/bulk_upload_column_change.py
@@ -7,32 +7,31 @@ from typing import Dict
 from pydantic import BaseModel, TypeAdapter
 from services.base.dynamo_service import DynamoDBService
 
-COLUMNS_TO_FETCH = "OdsCode,PdsOdsCode,UploaderOdsCode,ID"
-ATTRIBUTE_TO_REMOVE = "OdsCode"
-ATTRIBUTE_TO_ADD = "PdsOdsCode"
-UPLOADER_ODS_CODE = "PlaceHolder"
+TABLE = "ndrb_BulkUploadReport"
+COLUMNS_TO_FETCH = "FailureReason,ID"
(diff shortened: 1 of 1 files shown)


lambdas/scripts/bulk_upload_column_change.py
```diff
@@ -7,32 +7,31 @@
 from pydantic import BaseModel, TypeAdapter
 from services.base.dynamo_service import DynamoDBService
 
-COLUMNS_TO_FETCH = "OdsCode,PdsOdsCode,UploaderOdsCode,ID"
-ATTRIBUTE_TO_REMOVE = "OdsCode"
-ATTRIBUTE_TO_ADD = "PdsOdsCode"
-UPLOADER_ODS_CODE = "PlaceHolder"
+TABLE = "ndrb_BulkUploadReport"
```
MohammadIqbalAD-NHS, line 10: This is targeting a specific sandbox's table. Shouldn't this be modified dynamically depending on the environment? i.e `{env}_BulkUploadReport`

NogaNHS (approved):

lambdas/scripts/bulk_upload_column_change.py
AndyFlintNHS: I spoke with Noga and she's happy to keep it as-is as the script will be run infrequently. As a side note, it seems like {env} isn't an environment variable I could find in the codebase, it wouldn't be a trivial change from what I could see.

MohammadIqbalAD-NHS (approved):

PR #452, opened 2024-10-18
PRMP-1035 Remove duplicate stitch lambda call

steph-torres-nhs:


app/src/components/blocks/_lloydGeorge/lloydGeorgeViewRecordStage/LloydGeorgeViewRecordStage.tsx
```diff
@@ -209,17 +212,19 @@ function LloydGeorgeViewRecordStage({
             <PatientSimpleSummary />
             {!fullScreen ? (
                 <>
-                    {showMenu ? (
+                    <>
```
NogaNHS, line 215: remove

app/src/components/generic/pdfViewer/PdfViewer.tsx
```diff
@@ -5,7 +5,7 @@ type Props = { fileUrl: String };
 const PdfViewer = ({ fileUrl }: Props) => {
     useEffect(() => {
         const pdfObject = require('pdfobject');
-        pdfObject.embed(fileUrl + '#toolbar=0', '#pdf-viewer');
+        pdfObject.embed(fileUrl + '#toolbar', '#pdf-viewer');
```
NogaNHS, line 8: undo this change

app/src/components/generic/recordCard/RecordCard.tsx
```diff
@@ -26,27 +23,21 @@ function RecordCard({
...
     }, [refreshRecord]);
 
     const Record = () => {
-        if (isLoading) {
-            return (
-                <div className="pl-7">
-                    <ProgressBar status="Loading..." />
-                </div>
-            );
+        if (!cloudFrontUrl) {
```
AndyFlintNHS, line 39: Return null always jumps out to me, could this be replaced with:
`if (cloudFrontUrl) {
  return <PdfViewer fileUrl={cloudFrontUrl} />;
}`

app/src/components/generic/recordMenuCard/RecordMenuCard.test.tsx
```diff
@@ -49,6 +49,8 @@ const mockLinks: Array<LGRecordActionLink> = [
     },
 ];
 
+const className = 'lloydgeorge_record-stage_flex-row';
```
AndyFlintNHS, line 52: Nitpicky one, would it be worth referring to this as something that indicates that it's a formatting thing? Something about a variable called `className` jumps out to me.

AndyFlintNHS (changes requested):
Comments listed individually

app/src/pages/lloydGeorgeRecordPage/LloydGeorgeRecordPage.tsx
```diff
@@ -122,6 +124,7 @@ function LloydGeorgeRecordPage() {
                             numberOfFiles={numberOfFiles}
                             refreshRecord={refreshRecord}
                             cloudFrontUrl={cloudFrontUrl}
+                            showMenu={showMenu}
                         />
```
RioKnightleyNHS, line 128: Im not seeing any new unit tests for the LloydGeorgeRecordPage to reflect this new prop. Need tests to check what happens when this is toggled

app/src/components/generic/recordMenuCard/RecordMenuCard.tsx
```diff
@@ -15,33 +17,38 @@ type SubSectionProps = {
...
     const updateActions = recordLinks.filter((link) => link.type === RECORD_ACTION.UPDATE);
     const downloadActions = recordLinks.filter((link) => link.type === RECORD_ACTION.DOWNLOAD);
 
     if (recordLinks.length === 0) {
         return <></>;
     }
 
+    if (!showMenu) {
+        return null;
+    }
```
RioKnightleyNHS, line 30: Id say here to decide between returning null or a fragment so its consistent

app/src/components/generic/recordMenuCard/RecordMenuCard.test.tsx
RioKnightleyNHS: This classname can probably become a default prop of className in the actualy component, unless its overriden, so we dont need to paste a default one for every test

RioKnightleyNHS (changes requested):

app/src/pages/lloydGeorgeRecordPage/LloydGeorgeRecordPage.tsx
NogaNHS: there is a test for `Does not render the MenuCard if showMenu is false` in RecordMenuCard, do you think we need to tests the LloydGeorgeRecordPage ?

app/src/components/generic/recordMenuCard/RecordMenuCard.tsx
```diff
@@ -15,33 +17,38 @@ type SubSectionProps = {
     setStage: Dispatch<SetStateAction<LG_RECORD_STAGE>>;
 };
 
-function RecordMenuCard({ recordLinks, setStage }: Props) {
+function RecordMenuCard({ recordLinks, setStage, showMenu, className }: Props) {
     const updateActions = recordLinks.filter((link) => link.type === RECORD_ACTION.UPDATE);
```
RioKnightleyNHS, line 21: Add the default classname here in the props deconstruction props = 'classnamestring'. This means it still has a prop it can pass but if nothing is passed itll use this default string. It prevents needing to type it in the test cases.

RioKnightleyNHS (changes requested):

app/src/components/generic/recordCard/RecordCard.tsx
AndyFlintNHS: Discussed this together and came up with a ternary.

app/src/pages/lloydGeorgeRecordPage/LloydGeorgeRecordPage.tsx
RioKnightleyNHS: generally speaking we should but as there is already so much bloat in these tests we are probably fine

AndyFlintNHS (dismissed):

RioKnightleyNHS (dismissed):

abid-nhs (approved):

RioKnightleyNHS (approved):

NogaNHS (approved):

abid-nhs (approved):

PR #455, opened 2024-10-28
[PRMP-1063] blank bulk upload reports don't create a file

AndyFlintNHS:


abbas-khan10 (approved):

lambdas/services/bulk_upload_report_service.py
```diff
@@ -158,16 +158,19 @@ def generate_success_report(self, ods_reports: list[OdsReport]):
...
-            file_name=f"/tmp/{file_name}",
-        )
+            logger.info("Uploading daily success report file to S3")
+            self.s3_service.upload_file(
+                s3_bucket_name=self.reports_bucket,
+                file_key=f"{self.s3_key_prefix}/{file_name}",
+                file_name=f"/tmp/{file_name}",
+            )
+        else:
+            logger.info("No data to report for daily success report file")
```
NogaNHS, line 173: Is it possible to take this out to a new function?

AndyFlintNHS: It feels like scope creep to me as the only functionality that has changed properly is line 173, though I agree that the code would be cleaner this way. @abbas-khan10 what do you reckon?

abbas-khan10: I have no preference, I had considered this when first implementing as there is repetition. What were you thinking @NogaNHS? Something like:

```
def write_and_upload_additional_reports(file_name, headers, data_rows):
    // logic here
````

Comment (abbas-khan10) [reply lambdas/services/bulk_upload_report_service.py:173]: The bool would serve as a way of customising the log message for each report type as we want to be specific with which report wasn't created (open to discussion obviously)

Comment (AndyFlintNHS) [reply lambdas/services/bulk_upload_report_service.py:173]: As opposed to doing a true/false return, I was going to suggest something more like this:
```
        if data_rows:
            logger.info("Uploading daily success report file to S3")
            self.write_and_upload_additional_reports(file_name, headers, data_rows)
        else:
            logger.info("No data to report for daily success report file")

    def write_and_upload_additional_reports(self, file_name, headers, data_rows):
        self.write_additional_report_items_to_csv(
            file_name=file_name, headers=headers, rows_to_write=data_rows
        )

        self.s3_service.upload_file(
            s3_bucket_name=self.reports_bucket,
            file_key=f"{self.s3_key_prefix}/{file_name}",
            file_name=f"/tmp/{file_name}",
        )
```

File: lambdas/services/bulk_upload_report_service.py [hunk 2]
```diff
@@ -248,16 +233,11 @@ def generate_restricted_report(self, ods_reports: list[OdsReport]):
...
-        self.s3_service.upload_file(
-            s3_bucket_name=self.reports_bucket,
-            file_key=f"{self.s3_key_prefix}/{file_name}",
-            file_name=f"/tmp/{file_name}",
-        )
+        if data_rows:
+            logger.info("Uploading daily restricted report file to S3")
+            self.write_and_upload_additional_reports(file_name, headers, data_rows)
+        else:
+            logger.info("No data to report for daily deceased report file")
```
abbas-khan10, line 240: daily restricted*

lambdas/services/bulk_upload_report_service.py
```diff
@@ -400,3 +375,14 @@ def get_times_for_scan(self) -> tuple[datetime, datetime]:
         self.s3_key_prefix = f"bulk-upload-reports/{date_folder_name}"
 
         return start_timestamp, end_timestamp
+
+    def write_and_upload_additional_reports(self, file_name, headers, data_rows):
```
abbas-khan10, line 379: could we add param types here please? See above method params for examples

abbas-khan10 (dismissed):

lambdas/services/bulk_upload_report_service.py
```diff
@@ -400,3 +375,19 @@ def get_times_for_scan(self) -> tuple[datetime, datetime]:
         self.s3_key_prefix = f"bulk-upload-reports/{date_folder_name}"
 
         return start_timestamp, end_timestamp
+
+    def write_and_upload_additional_reports(
+        self,
+        file_name: str,
+        headers: list[str],
+        data_rows: list[list[str]],
+    ):
```
NogaNHS, line 384: Can you add a small unit test for this?

lambdas/services/bulk_upload_report_service.py
AndyFlintNHS: Corrected

lambdas/services/bulk_upload_report_service.py
AndyFlintNHS: Logic has been broken out into a new function

NogaNHS (approved):

abbas-khan10 (approved):

PR #457, opened 2024-10-30
[PRMP-1064] - Update bulk upload summary and ODS reports to contain t…

abid-nhs:
…otal ingested

### DIFF
--- a/lambdas/models/bulk_upload_report_output.py
+++ b/lambdas/models/bulk_upload_report_output.py
@@ -19,7 +19,7 @@ class ReportBase:
         self.total_suspended = set()
         self.total_deceased = set()
         self.total_restricted = set()
-        self.total_ingested = 0
+        self.total_ingested = set()
 
     def get_total_successful_nhs_numbers(self) -> list:
         if self.total_successful:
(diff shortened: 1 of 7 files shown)


abbas-khan10 (dismissed):

lambdas/models/bulk_upload_report_output.py
```diff
@@ -70,6 +71,7 @@ def populate_report(self):
                 self.process_failed_report_item(item)
 
         self.set_unique_failures()
+        self.set_total_ingested()
```
abbas-khan10, line 74: Is there no way we can just do the count of total ingested in the for loop above? Just store each `item.nhs_number` in the ReportBase total_ingested as a set() rather than an int. This way it will only count unique nhs numbers. Then in the summary, similar to the other calculations we do `total_ingested=summary_report.get_total_ingested_count()`?

lambdas/models/bulk_upload_report_output.py
```diff
@@ -19,6 +19,7 @@ def __init__(
         self.total_suspended = set()
         self.total_deceased = set()
         self.total_restricted = set()
+        self.total_ingested = 0
```
RioKnightleyNHS, line 23: I think we would want to follow one standard here with set() and update(). Or making it dynamic in some way

RioKnightleyNHS (changes requested):

lambdas/models/bulk_upload_report_output.py
abbas-khan10: `self.total_ingested.add(item.nhs_number)`

abbas-khan10 (approved):

RioKnightleyNHS (approved):

PR #458, opened 2024-11-04
PRMP-1066 add RegisteredAtUploaderPractice flag to success and rejected bulk upload reports

steph-torres-nhs:


lambdas/models/bulk_upload_report_output.py
```diff
@@ -77,20 +77,30 @@ def populate_report(self):
         self.set_unique_failures()
 
     def process_successful_report_item(self, item: BulkUploadReport):
-        self.total_successful.add((item.nhs_number, item.date))
 
         if item.pds_ods_code == PatientOdsInactiveStatus.SUSPENDED:
             self.total_suspended.add((item.nhs_number, item.date))
+
```
abbas-khan10, line 83: Can we remove these break lines?

lambdas/models/bulk_upload_report.py
```diff
@@ -28,6 +29,19 @@ class BulkUploadReport(BaseModel):
...
+        if self.pds_ods_code in PatientOdsInactiveStatus.list():
+            return self.pds_ods_code
+
+        elif (
+            self.uploader_ods_code != self.pds_ods_code
+            and self.pds_ods_code not in PatientOdsInactiveStatus.list()
+        ):
+            return False
+
+        return True
```
abbas-khan10, line 43: Can we not simplify this logic?

```
def get_registered_at_uploader_practice_status(self):
    if self.pds_ods_code in PatientOdsInactiveStatus.list():
        return self.pds_ods_code
    return self.uploader_ods_code == self.pds_ods_code
```

lambdas/tests/unit/models/test_bulk_upload_report.py
```diff
@@ -92,3 +104,20 @@ def test_failed_upload_ids_and_timestamp_are_auto_populated_if_not_given(mock_uu
     ).model_dump(by_alias=True, exclude_none=True)
 
     assert actual == expected
+
+
+def test_get_registered_at_uploader_status_returns_true_uploader_ods_and_pds_ods_are_the_same():
+
```
abbas-khan10, line 110: remove break lines after method declarations please

lambdas/models/bulk_upload_report.py
```diff
@@ -28,6 +29,11 @@ class BulkUploadReport(BaseModel):
     uploader_ods_code: str = Field(alias=MetadataReport.UploaderOdsCode)
     reason: Optional[str] = Field(default="", alias=MetadataReport.Reason)
 
+    def get_registered_at_uploader_practice_status(self):
```
abbas-khan10, line 32: Sorry forgot to add this as part of the last review, could we also add a return type for this function as it will be `-> str` and then cast the response of the boolean `self.uploader_ods_code == self.pds_ods_code` to return as a string value also? Just so it's clear whats being done here as we've returning two different data types from a glance. somethging like this?
```
def get_registered_at_uploader_practice_status(self) -> str:
    if self.pds_ods_code in PatientOdsInactiveStatus.list():
        return self.pds_ods_code
    return str(self.uploader_ods_code == self.pds_ods_code)
```

lambdas/models/bulk_upload_report.py
```diff
@@ -28,6 +29,11 @@ class BulkUploadReport(BaseModel):
     uploader_ods_code: str = Field(alias=MetadataReport.UploaderOdsCode)
     reason: Optional[str] = Field(default="", alias=MetadataReport.Reason)
 
+    def get_registered_at_uploader_practice_status(self) -> str:
+        if self.pds_ods_code in PatientOdsInactiveStatus.list():
```
AndyFlintNHS, line 33: We can convert this into a ternary statement:
`return self.pds_ods_code if self.pds_ods_code in PatientOdsInactiveStatus.list() else str(self.uploader_ods_code == self.pds_ods_code)`

It's quite a long line of code, I don't know how this'd be default-formatted by Python but for readability (at least from a Java perspective 😄 it'd be nice like this:

```
return self.pds_ods_code 
    if self.pds_ods_code in PatientOdsInactiveStatus.list() 
    else str(self.uploader_ods_code == self.pds_ods_code)
```

lambdas/tests/unit/helpers/data/bulk_upload/expected_success_report.csv
```diff
@@ -1,11 +1,11 @@
...
+9000000000,Y12345,2012-01-13,SUSP
+9000000001,Y12345,2012-01-13,DECE
+9000000002,Y12345,2012-01-13,REST
+9000000003,Y12345,2012-01-13,True
+9000000004,Y12345,2012-01-13,False
+9000000009,Z12345,2012-01-13,SUSP
+9000000010,Z12345,2012-01-13,DECE
+9000000011,Z12345,2012-01-13,REST
+9000000012,Z12345,2012-01-13,False
+9000000013,Z12345,2012-01-13,True
```
AndyFlintNHS, line 11: Just going to have to trust that these are all correct, I'm assuming this has come from running the unit tests -> failed -> here's what it expected and you've dropped it in?

lambdas/tests/unit/services/test_bulk_upload_report_service.py
```diff
@@ -439,13 +439,21 @@ def test_generate_summary_report_with_two_ods_reports(
...
+        f"daily_statistical_report_bulk_upload_success_{MOCK_TIMESTAMP}.csv"
+    )
     bulk_upload_report_service.write_and_upload_additional_reports = mocker.MagicMock()
 
     test_ods_reports = bulk_upload_report_service.generate_ods_reports(
         MOCK_REPORT_ITEMS_ALL,
     )
 
     bulk_upload_report_service.generate_success_report(test_ods_reports)
+    expected = readfile("expected_success_report.csv")
```
AndyFlintNHS, line 452: Isn't this the logic that's being tested below in the test on line 588? Is there another file being created here? We don't want to test the logic that's done as part of `write_and_upload_additional_reports()`, we just want to test it's being called, I just pulled that out into a separate test last week.

AndyFlintNHS: Discussed this on a call - Steph & Abid's approach avoids a much more cumbersome & verbose refactor. The existing logic did not check what was actually in the csv that was created.

AndyFlintNHS (approved):

abbas-khan10 (approved):

AndyFlintNHS (dismissed):

lambdas/enums/metadata_report.py
```diff
@@ -11,6 +11,7 @@ class MetadataReport(StrEnum):
     Date = "Date"
     Timestamp = "Timestamp"
     ID = "ID"
+    RegisteredAtUploaderPractice = "RegisteredAtUploaderPractice"
 
     @staticmethod
     def list():
```
abid-nhs, line 17: This method is no longer being used anywhere in the codebase. it can be removed.

abid-nhs (changes requested):

lambdas/enums/metadata_report.py
abid-nhs: This has been actioned by Steph

AndyFlintNHS (approved):

abid-nhs (approved):

PR #462, opened 2024-11-11
PRMP-1122 - Add new records to NRL SQS 

jack-nhs:


lambdas/repositories/bulk_upload/bulk_upload_sqs_repository.py
```diff
@@ -39,3 +40,6 @@ def put_sqs_message_back_to_queue(self, sqs_message: dict):
             message_body=sqs_message["body"],
             nhs_number=nhs_number,
         )
+
+    def send_message_to_nrl_fifo(self, queue_url: str, message: NrlSqsMessage):
+        self.sqs_repository.send_message_standard(queue_url, message.model_dump())
```
NogaNHS, line 45: The queue we're using is a FIFO, group id is required for FIFO queues.

lambdas/tests/unit/services/test_bulk_upload_service.py
```diff
@@ -200,7 +201,10 @@ def test_handle_sqs_message_happy_path(
     mock_pds_validation,
     mock_ods_validation,
 ):
+    # given
```
AndyFlintNHS, line 204: Apologies, I know I added this in as an example when we paired, but for consistency can we remove the given, when, then comments please?

lambdas/enums/snomed_codes.py
```diff
@@ -0,0 +1,9 @@
+from enum import StrEnum
+
+
+class SnomedCodesType(StrEnum):
+    LLOYD_GEORGE = "16521000000101"
```
AndyFlintNHS, line 5: Are we planning on adding more of these? Similar to the previous comment, if there aren't going to be any more codes we should look at making this a constant somewhere instead of a single-value Enum. If we're adding more in future, I'm happy for this to stay.

AndyFlintNHS (changes requested):

PR #464, opened 2024-11-12
PRMP-1069: Fix metadata nhs number validation failure 

abbas-khan10:


lambdas/models/staging_metadata.py
```diff
@@ -51,15 +48,14 @@ def ensure_gp_practice_code_non_empty(
...
 
     @field_validator("nhs_number")
     @classmethod
     def validate_nhs_number(cls, nhs_number: str) -> str:
-        if nhs_number.isdigit() and len(nhs_number) == 10:
+        if nhs_number and nhs_number.isdigit():
             return nhs_number
 
-        raise ValueError("NHS number must be a 10 digit number")
+        return NHS_NUMBER_PLACEHOLDER
```
steph-torres-nhs, line 61: Do we want a place holder, may end up with a lot of files for the same NHS number in the bulkupload table, could potentially skew reporting and make differentiating files associated to different patients difficult?
 BulkUpload lambda could handle this?

abbas-khan10: We need some value here to pass to the SQS queue as the nhs number is used as the key in the message dictionary. This solution will be refined in future tickets, for now the placeholder will only be used for missing/non-numerical nhs numbers just so ingestion can continue with invalid NHS numbers. This will be caught and rejected in the bulk upload lambda, and we'll be able to identify these if the nhs number matches the placeholder value

steph-torres-nhs (approved):

AndyFlintNHS (approved):

PR #465, opened 2024-11-14
PRMP-1192 - delete pointer

NogaNHS:


lambdas/services/document_deletion_service.py
```diff
@@ -77,3 +84,18 @@ def delete_specific_doc_type(
                 {"Results": "Failed to delete documents"},
             )
             raise DocumentDeletionServiceException(500, LambdaError.DocDelClient)
+
+    def send_sqs_message_to_remove_pointer(self, nhs_number: str):
+        delete_nrl_message = NrlSqsMessage(
+            nhs_number=nhs_number,
+            action="delete",
+            snomed_code_doc_type="16521000000101",
```
ollieb-madetech, line 92: These could/should be ENUMS - do we currently have ENUMS for these?

lambdas/services/document_deletion_service.py
```diff
@@ -77,3 +84,18 @@ def delete_specific_doc_type(
...
+            snomed_code_category="734163000",
+        )
+        sqs_group_id = f"NRL_delete_{uuid.uuid4()}"
```
steph-torres-nhs, line 95: does order matter here or is this because fifo queues require group id?

lambdas/handlers/manage_nrl_pointer_handler.py
```diff
@@ -50,14 +46,21 @@ def lambda_handler(event, context):
             nrl_verified_message = nrl_message.model_dump(
                 by_alias=True, exclude_none=True, exclude_defaults=True
             )
-            document = (
-                FhirDocumentReference(
-                    **nrl_verified_message, custodian=nrl_api_service.end_user_ods_code
+            if nrl_message.action == "create":
```
AndyFlintNHS, line 49: Would this `if/elif` be more appropriate as a switch statement?

oliverbeumkes-nhs (approved):

AndyFlintNHS (approved):
Pair reviewed with Ollie - happy with the changes made.

robg-nhs (dismissed):
Question around how this will interact with Prod

PR #470, opened 2024-11-27
PRMP-1188 Create a lambda to handle MNS notifications

steph-torres-nhs:


lambdas/handlers/mns_notification_handler.py
```diff
@@ -0,0 +1,44 @@
...
+            sqs_message = json.loads(sqs_message["body"])
+
+            mns_message = MNSSQSMessage(**sqs_message)
+            MNSSQSMessage.model_validate(mns_message)
+
+            if mns_message.type in MNSNotificationTypes.__members__.values():
+                notification_service = MNSNotificationService()
+                notification_service.handle_mns_notification(mns_message)
+                logger.info("Continuing to next message.")
+        except Exception as error:
```
abbas-khan10, line 42: Do we know what specific exception can be raised here?

steph-torres-nhs: There are a few different types that could bubble up and just wanted to make sure they are caught. Is there a better way to handle this?

steph-torres-nhs: Thought about this, there could be a pydantic validation error here that doesn't get handled, something like malformed message we have else where in the code base..?

abbas-khan10: I guess a catch all is required if there's a risk of this falling over, I'm not against the Exception but if we can catch what we know that would be ideal

lambdas/tests/unit/services/test_process_mns_message_service.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+gp_change_message = MNSSQSMessage(**MOCK_GP_CHANGE_MESSAGE_BODY)
+death_notification_message = MNSSQSMessage(**MOCK_DEATH_MESSAGE_BODY)
+informal_death_notification_message = MNSSQSMessage(**MOCK_INFORMAL_DEATH_MESSAGE_BODY)
+removed_death_notification_message = MNSSQSMessage(**MOCK_REMOVED_DEATH_MESSAGE_BODY)
+
+
+def test_handle_gp_change_message_called_message_type_gp_change(mns_service, mocker):
+    mocker.patch.object(mns_service, "handle_gp_change_notification")
```
abbas-khan10, line 38: `handle_gp_change_notification` is used multiple times in this file, any way we can create a fixture and also `assert_called_with` instead of just an `assert_called()`

.github/workflows/base-lambdas-reusable-deploy-all.yml
```diff
@@ -402,4 +402,17 @@ jobs:
     secrets:
       AWS_ASSUME_ROLE: ${{ secrets.AWS_ASSUME_ROLE }}
 
+  deploy_mns_notification_lambda:
```
NogaNHS, line 405: Can you add a condition for when the env is a sandbox?

lambdas/handlers/mns_notification_handler.py
```diff
@@ -0,0 +1,44 @@
...
+        try:
+            sqs_message = json.loads(sqs_message["body"])
+
+            mns_message = MNSSQSMessage(**sqs_message)
+            MNSSQSMessage.model_validate(mns_message)
+
+            if mns_message.type in MNSNotificationTypes.__members__.values():
+                notification_service = MNSNotificationService()
+                notification_service.handle_mns_notification(mns_message)
+                logger.info("Continuing to next message.")
```
NogaNHS, line 41: Can move this log to line 44 outside of the try-catch

lambdas/handlers/mns_notification_handler.py
```diff
@@ -0,0 +1,44 @@
...
+
+    for sqs_message in sqs_messages:
+        try:
+            sqs_message = json.loads(sqs_message["body"])
+
+            mns_message = MNSSQSMessage(**sqs_message)
+            MNSSQSMessage.model_validate(mns_message)
+
+            if mns_message.type in MNSNotificationTypes.__members__.values():
+                notification_service = MNSNotificationService()
```
NogaNHS, line 39: Do we want a new instance for every message?

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+from utils.audit_logging_setup import LoggingService
+from utils.exceptions import PdsErrorException
+from utils.utilities import get_pds_service
+
+logger = LoggingService(__name__)
+
+
+class MNSNotificationService:
+    def __init__(self):
+        self.message = None
```
NogaNHS, line 19: Should the message be a property of this service?

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+        self.dynamo_service = DynamoDBService()
+        self.table = os.getenv("LLOYD_GEORGE_DYNAMODB_NAME")
+        self.pds_service = get_pds_service()
+        self.sqs_service = SQSService()
+        self.queue = os.getenv("MNS_NOTIFICATION_QUEUE_URL")
+
+    def handle_mns_notification(self, message: MNSSQSMessage):
+        try:
+            if message.type == MNSNotificationTypes.CHANGE_OF_GP.value:
```
NogaNHS, line 28: no need for .value with strEnum

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+                logger.info("Handling GP change notification.")
+                self.handle_gp_change_notification(message)
+
+            elif message.type == MNSNotificationTypes.DEATH_NOTIFICATION.value:
+                logger.info("Handling death status notification.")
+                self.handle_death_notification(message)
+
+        except PdsErrorException:
+            self.send_message_back_to_queue(message)
```
NogaNHS, line 37: Maybe add a log saying we're putting the message back?

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+
+        except Exception as e:
```
NogaNHS, line 39: I think we should catching a general Exception only in one place. 
So can you change this to a more specific one?

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+            logger.info(
+                f"Unable to process message: {message.id}, of type: {message.type}"
+            )
+            logger.info(f"{e}")
+            return
```
NogaNHS, line 44: needed?

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+            )
+            return
+
+        try:
+            patient_documents = self.get_patient_documents(message.subject.nhs_number)
+
+            if len(patient_documents) < 1:
+                logger.info("Patient is not held in the National Document Repository.")
+                logger.info("Moving on to the next message.")
+                return
```
NogaNHS, line 81: We're repeating this logic from `handle_gp_change_notification`

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+
+            if (
+                message.data["deathNotificationStatus"]
+                == DeathNotificationStatus.REMOVED.value
+            ):
+                update_ods_code = self.get_updated_gp_ods(message.subject.nhs_number)
+                self.update_patient_details(patient_documents, update_ods_code)
+                logger.info("Update complete for death notification change.")
+        except ClientError as e:
+            logger.error(f"{e}")
```
NogaNHS, line 101: Any reason we're catching `ClientError` here and not in `handle_gp_change_notification`? Maybe best to move this into `handle_mns_notification`?

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def is_informal_death_notification(self, message: MNSSQSMessage) -> bool:
+        return (
+            message.data["deathNotificationStatus"]
+            == DeathNotificationStatus.INFORMAL.value
+        )
```
NogaNHS, line 107: Any reason this a separate function but the other checks for death status are not?

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def get_patient_documents(self, nhs_number: str):
+        logger.info("Getting patient document references...")
+        response = self.dynamo_service.query_table_by_index(
+            table_name=self.table,
+            index_name="NhsNumberIndex",
+            search_key="NhsNumber",
+            search_condition=nhs_number,
+        )
```
NogaNHS, line 116: Maybe also filter the record wasn't deleted? Do we want to update if it was?

lambdas/handlers/mns_notification_handler.py
```diff
@@ -0,0 +1,44 @@
...
+
+    sqs_messages = event.get("Records", [])
+
+    for sqs_message in sqs_messages:
+        try:
+            sqs_message = json.loads(sqs_message["body"])
+
+            mns_message = MNSSQSMessage(**sqs_message)
+            MNSSQSMessage.model_validate(mns_message)
+
```
NogaNHS, line 37:
```suggestion
request_context.patient_nhs_no = message.subject.nhs_number
```

.github/workflows/subscribe-to-mns.yml
```diff
@@ -16,18 +16,17 @@ on:
       AWS_ASSUME_ROLE:
         required: true
 permissions:
-  pull-requests: write
   id-token: write # This is required for requesting the JWT
   contents: read # This is required for actions/checkout
 jobs:
-  batch_update_build_docker_image:
+  placeholder:
```
abbas-khan10, line 22: Do we still want this to be a placeholder?

steph-torres-nhs: the workflow that we want is currently on a different ticket that hasn't yet been merged in.

lambdas/enums/mns_notification_types.py
```diff
@@ -0,0 +1,6 @@
+from enum import StrEnum
+
+
+class MNSNotificationTypes(StrEnum):
+    CHANGE_OF_GP = "pds-change-of-gp-1"
```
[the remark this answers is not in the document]
steph-torres-nhs: these are from the MNS docs linked in the ticket

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+            logger.info(f"{e}")
+
+    def handle_gp_change_notification(self, message: MNSSQSMessage):
+        patient_document_references = self.get_patient_documents(
+            message.subject.nhs_number
+        )
+
+        if not self.patient_is_present_in_ndr(patient_document_references):
+            return
```
AndyFlintNHS, line 53: Given that both the change of GP and the death notifications (and I'd imagine any potential future notifications) involve this step of pulling out the document references and then checking for the presence of this patient in the NDR, I'd suggest we pull these lines out and do this check on line 29 before the `if/elif` to avoid duplication and clean these steps up.

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,150 @@
...
+                        DocumentReferenceMetadataFields.LAST_UPDATED.value: int(
+                            datetime.now().timestamp()
+                        ),
+                    },
+                )
+
+        logger.info("Update complete for change of GP")
+
+    def handle_death_notification(self, message: MNSSQSMessage):
+        if self.is_informal_death_notification(message):
```
AndyFlintNHS, line 73: Similar to the above feedback, I'd suggest getting rid of the `is_informal_death_notification` and `is_formal_death_notification` style functions and just changing this to a `match/case` statement on `message.data["deathNotificationStatus"]` for conciseness. We've got a bunch of statements in the style of
```
if X function is true -> return
if Y function is true -> do this
if Z function is true -> do this
```

And I think we could simplify this to just be (very pseudo-code-y sorry!)
```
Match message.data["deathNotificationStatus"]
    case  DeathNotificationStatus.INFORMAL -> do this
    case  DeathNotificationStatus.FORMAL -> do this
    case DeathNotificationStatus.REMOVED -> do this
``` 

This, with the above feedback means that we could massively simplify the `handle_death_notification` function. Please drop me a message if you'd like any clarification.

lambdas/services/process_mns_message_service.py
AndyFlintNHS: @steph-torres-nhs was this addressed? It still looks like it's not filtering from what I can see

lambdas/services/process_mns_message_service.py
```diff
@@ -0,0 +1,150 @@
...
+        logger.info("Getting patient document references...")
+        response = self.dynamo_service.query_table_by_index(
+            table_name=self.table,
+            index_name="NhsNumberIndex",
+            search_key="NhsNumber",
+            search_condition=nhs_number,
+        )
+        return response["Items"]
+
+    def update_patient_details(self, patient_documents: list[dict], code: str) -> None:
```
AndyFlintNHS, line 120: Can we rename this to be `update_patient_ods_code`? It's not all their details.

lambdas/services/process_mns_message_service.py
AndyFlintNHS: Is this queue working in a different way? I've never had to put anything back on a message queue manually, it's always been done on an acknowledgement system. `If the message isn't acknowledged -> retry after a set amount of time.` is the system I'm used to.

AndyFlintNHS (changes requested):

lambdas/services/process_mns_message_service.py
steph-torres-nhs: Yeah this was addressed after a discussion with Ifey, keeping as is for time being

AndyFlintNHS (approved):

PR #479, opened 2024-12-04
[PRMP-1257] Add log messages to adhere to audit requirements

steph-torres-nhs:


lambdas/handlers/manage_nrl_pointer_handler.py
```diff
@@ -58,7 +59,16 @@ def lambda_handler(event, context):
                         .json()
                     )
 
+                    logger.info(
```
AndyFlintNHS, line 62: I notice that the ticket mentions '_When a pointer is created **two** Audit Logs must be made in the system_'

I'll have a chat with Rob (he made the ticket) and work out if there's an intended format for this, but I'd assume we want to create a JSON object and log that out.

MohammadIqbalAD-NHS (approved):
Generally happy with the PR. There are some very minor suggestions regarding consistent logging i.e some places `HTTPVerb:`, others `HTTP Verb:`, same with `Nhs Number` and `Nhs number`. Feel free to address if deemed appropriate, I am not too fussed with it.

lambdas/models/nrl_sqs_message.py
```diff
@@ -6,13 +6,13 @@
...
+        alias_generator=AliasGenerator(serialization_alias=to_camel),
+        use_enum_values=True,
+    )
     content_type: str = "application/pdf"
     language: str = "en-UK"
-    url: str = ""
-    size: int = 0
-    hash: str = ""
-    title: str = ""
-    creation: str = ""
```
NogaNHS, line 15: Change defaults to none instead

PR #487, opened 2024-12-12
PRMP-1342: Create lambda to handle deleted document references

abbas-khan10:


lambdas/handlers/delete_document_object_handler.py
```diff
@@ -0,0 +1,59 @@
...
+    except (ValueError, ValidationError) as e:
+        logger.error(
+            f"Failed to parse Document Reference from deleted record: {str(e)}",
+            {"Results": "Failed to delete document"},
+        )
+        raise DocumentDeletionServiceException(
+            400, LambdaError.DynamoInvalidStreamEvent
+        )
+
+    return ApiGatewayResponse(
```
steph-torres-nhs, line 57: gateway integration set to false in terraform, do we need it for returning gateway responses?

jack-nhs (approved):

PR #491, opened 2024-12-17
[PRMP-1330] - Add service-updates-link to footer

MohammadIqbalAD-NHS:


oliverbeumkes-nhs (dismissed):

app/src/components/generic/serviceUpdatesLink/ServiceUpdatesLink.tsx
```diff
@@ -0,0 +1,17 @@
+import { Footer as NHSFooter } from 'nhsuk-react-components';
+
+const ServiceUpdatesLink = () => {
+    return (
+        <NHSFooter.ListItem
+            href="https://digital.nhs.uk/services/access-and-store-digital-patient-documents/service-updates"
+            data-testid="service-updates-link"
+            rel="noopener"
+            target="_blank"
+            aria-label="(Service updates - Opens in a new tab)"
```
steph-torres-nhs, line 10: Don't think we want brackets in the aria-label, think some screen readers read them , believe we have in mind to remove the brackets from the other link in the future...?

MohammadIqbalAD-NHS: Good point! Has now been addressed.

steph-torres-nhs (approved):

app/src/components/layout/footer/Footer.tsx
```diff
@@ -15,6 +16,7 @@ function Footer() {
                 >
                     Privacy notice
                 </NHSFooter.ListItem>
+                <ServiceUpdatesLink />
```
abid-nhs, line 19: As discussed I was wondering if it would be worth making the <NHSFooter.ListItem for ServiceUpdatesLink in the same file rather than separate one like the privacy link Footer list item. This might help with consistency especially since the ServiceUpdatesLink might not be usable in other places as its made as a NHSFooter.listitem

abid-nhs (changes requested):
It would be worthwhile modifying Footer.test.txs to have the first test/it block to check all the links are present

app/src/components/layout/footer/Footer.tsx
MohammadIqbalAD-NHS: I have modified this to no longer use a component but rather a constant within the Footer.tsx

abid-nhs (approved):

abid-nhs (changes requested):
looks good however cypress tests seems to be failing

abid-nhs (approved):

PR #492, opened 2024-12-17
[PRMP-1240] Reduce width of NHS number input field

mark-start-nhs:
for

https://nhse-dsic.atlassian.net/browse/PRMP-1240 

className="nhsuk-input--width-10"    is applied

the next smallest is     nhsuk-input--width-5

which is awkwardly short 

className="nhsuk-input--width-10" :

image


app/src/pages/patientSearchPage/PatientSearchPage.tsx
```diff
@@ -115,6 +115,7 @@ function PatientSearchPage() {
                     <Input
                         id="nhs-number-input"
                         data-testid="nhs-number-input"
+                        className="nhsuk-input--width-10"
```
abid-nhs, line 118: Would it be worth having something like `maxlength="12"` limiting it to 12 chars

abid-nhs (approved):

steph-torres-nhs (approved):

PR #494, opened 2024-12-17
[PRMP-1237] Underline nav bar links in the nav bar

mark-start-nhs:
for

https://nhse-dsic.atlassian.net/browse/PRMP-1237

### DIFF
--- a/app/src/styles/App.scss
+++ b/app/src/styles/App.scss
@@ -678,9 +678,16 @@ $hunit: '%';
     }
 
     &_item {
-        a,
-        a:hover {
+
+        a {
             color: $color_nhsuk-white;
+            text-decoration: underline;
+            text-decoration-color: $color_nhsuk-white;
+            text-underline-offset: 2px;
+        }
+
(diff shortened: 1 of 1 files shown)


oliverbeumkes-nhs (changes requested):
This change is not required and is an override to the NHS.UK Frontend package nav bar which has been accounted fr in the update from 7.0.0 to 9.1.0. 

image

We had agreed in refinement last week not to override these methods until at least the ticket PRMP-1348 had gone in so that not to override future Design changes the will come in from the NHS.UK Frontend.

The screenshot is of this issue being fixed by updating the package with any additional css changes from our end.

oliverbeumkes-nhs:
We @abid-nhs  and @MohammadIqbalAD-NHS have covered this package upgrade and it is going to take some further consideration due to changes to the header and footer at the very least. This will be fine for a quick fix and has been noted on the PRMP-1348 ticket.

abid-nhs (approved):

PR #495, opened 2024-12-17
[PRMP-1239] NEW SERVICE updated to be New service

mark-start-nhs:
for 

https://nhse-dsic.atlassian.net/browse/PRMP-1239

### DIFF
--- a/app/cypress/e2e/0-ndr-core-tests/homepage.cy.js
+++ b/app/cypress/e2e/0-ndr-core-tests/homepage.cy.js
@@ -34,7 +34,7 @@ describe('Home Page', () => {
     });
 
     it('displays service banner', { tags: 'regression' }, () => {
-        cy.get('.govuk-phase-banner__content__tag').should('have.text', 'New Service');
+        cy.get('.govuk-phase-banner__content__tag').should('have.text', 'New service');
         cy.get('.govuk-phase-banner__text').should(
             'have.text',
             'Your feedback will help us to improve this service.',
--- a/app/src/styles/App.scss
+++ b/app/src/styles/App.scss
@@ -793,6 +793,10 @@ $hunit: '%';
     border: none;
 }
 
+.govuk-phase-banner__content__tag {
+    text-transform: none;
+}
+
 .govuk-warning-text__icon {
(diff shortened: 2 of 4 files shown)


app/src/styles/App.scss
```diff
@@ -793,6 +793,10 @@ $hunit: '%';
     border: none;
 }
 
+.govuk-phase-banner__content__tag {
```
oliverbeumkes-nhs, line 796: With reference to the above suggestion when using the `nhsuk-tag` we match he NHS style so this can be removed as i doesn't apply any text transformation so this can be removed. That being said `nhsuk-tag` doesn't have the `margin-right: 10px` so we could add that to the tag or add `margin-left: 10px` to the `govuk-phase-banner__text`.

oliverbeumkes-nhs (changes requested):
Please review the comments above, I have requested these changes for the following reason: By keeping inline with the NHS.UK frontend it protects us when future styling comes in with the NHS.UK frontend package to keep us inline with the Design's (as overriding them can cause issues down the line).

That being said we also have a ticket to update the current NHS.UK frontend package PRMP-1348, which was to be done before overriding any styles as would fix a lot of the changes.

oliverbeumkes-nhs:
We @abid-nhs  and @MohammadIqbalAD-NHS have covered the package upgrade and it is going to take some further consideration due to changes to the header and footer at the very least. This will be fine for a quick fix and has been noted on the PRMP-1348 ticket.

app/cypress/e2e/0-ndr-core-tests/homepage.cy.js
```diff
@@ -34,7 +34,7 @@ describe('Home Page', () => {
     });
 
     it('displays service banner', { tags: 'regression' }, () => {
-        cy.get('.govuk-phase-banner__content__tag').should('have.text', 'New Service');
+        cy.get('.govuk-phase-banner__content__tag').should('have.text', 'New service');
```
abid-nhs, line 37: instead of .govuk-phase-banner__content__tag can we use `nhsuk-tag`

abid-nhs (changes requested):

abid-nhs (approved):

PR #506, opened 2025-01-09
[PRMP-1421] Add missing repository name in checkout action in workflow

NogaNHS:


chrisbloe-nhse (approved):
Seems reasonable. Ensures the correct repo is checked out when reusing the workflow externally.

May need a wider review to check for other possible occurrences, but happy with this immediate change for now.

MohammadIqbalAD-NHS (approved):

PR #513, opened 2025-01-17
[PRMP-1363] New footer link - Help and guidance

jack-nhs:


AndyFlintNHS (approved):

steph-torres-nhs (changes requested):
Would we be able to add a test verifying the component renders following other footer link pattern?

steph-torres-nhs (approved):

PR #514, opened 2025-01-19
[PRMP-1372] New PDF viewer (pdf.js)

mark-start-nhs:
add pdfjs setup script + custom_css to apply,
add iframe of pdfjs to react app,
add pdfjs to nginx.conf + remove some of the deprecated request permission-policies 
update tests to point to the new iframe , <-- there needs to be a new test to check the iframe contents 
update .gitignore to keep the pdfjs build but not the setup_pdfjs.py temp files,
add gulp-cli as a dev dependency used to compile pdfjs


app/pdfjs/setup_pdfjs.py
```diff
@@ -0,0 +1,143 @@
...
+    finally:
+
+        print_message( "finally" )
+        delete_temp_dir()
+        print_message("PDF.js setup complete", newline_before=True, newline_after=True)
+
+    sys.exit(0)
+
+
+
```
AndyFlintNHS, line 141: Given this is just a script that'll be run once in a while it's not enough that I'd block this, but there's a lot of whitespace in this script that we could clean up in a future iteration.

AndyFlintNHS (approved):
Overall, it’s a really solid chunk of work I’m happy to go ahead with a few considerations:

1. That we’re happy that the print/rotate buttons for the PDF are hidden in a menu
2. That permitting 'unsafe-inline' is okay within the nginx.conf Content-Security-Policy
3. I worry about maintainability, this approach means that we will have full customisability of our PDF viewer, at the cost of having to run Mark’s `setup_pdfjs.py` script to update versions and add the entire pdf.js code into our codebase.

app/pdfjs/setup_pdfjs.py
mark-start-nhs: can definitely delete the white space in the next iteration

PR #520, opened 2025-01-22
[PRMP-1456] - Add workflow permissions for Sonarcloud

danielleandersonNHS:


chrisbloe-nhse (approved):
Yes, we need these permissions.

PR #532, opened 2025-02-04
[PRMP-1502] Remove 400 status codes and update if necessary

steph-torres-nhs:


lambdas/handlers/nrl_get_document_reference_handler.py
```diff
@@ -25,10 +25,6 @@
 )
 def lambda_handler(event, context):
     try:
-        if not event:
-            raise NRLGetDocumentReferenceException(
-                400, LambdaError.DocumentReferenceInvalidRequest
-            )
```
NogaNHS, line 28: Why are we removing this and not returning 403 instead?

NogaNHS: Or a 500?
We should return a OperationOutcome

steph-torres-nhs: after discussion putting back in, as it is helpful for us when we manually trigger the lambda even if the API will never trigger the lambda without an event.

jack-nhs (approved):

AndyFlintNHS (approved):

PR #533, opened 2025-02-04
[PRMP-1503] Fix bug when deploying lambda edge versions

abbas-khan10:


chrisbloe-nhse (approved):
Adding `--publish` generates the new version for CloudFront to point to.

danielleandersonNHS (approved):

PR #538, opened 2025-02-07
[PRMP-1518] Hide server version

steph-torres-nhs:
change config file to hide the server version

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -11,7 +11,7 @@ RUN npm run build
 
 # Host the App
 FROM nginx:latest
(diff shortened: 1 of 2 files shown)


chrisbloe-nhse (changes requested):
I don't think this comment is relevant without the other change, so let's change it to indicate the header will not include the version.

jack-nhs (approved):

MohammadIqbalAD-NHS (approved):

chrisbloe-nhse (approved):
Yup, let's give this a go - it should remove the server version :).

PR #539, opened 2025-02-10
[PRMP-1517] Change to auth service checks

NogaNHS:


lambdas/services/authoriser_service.py
```diff
@@ -83,6 +91,9 @@ def find_login_session(self, ndr_session_id):
 
         try:
             current_session = query_response["Items"][0]
+            list_of_allowed_nhs_numbers = current_session.get("AllowedNHSNumbers", [])
```
steph-torres-nhs, line 94: should this default to an empty string instead given allowedNHSNumbers is a string..?

lambdas/services/search_patient_details_service.py
```diff
@@ -93,3 +105,44 @@ def check_if_user_authorise(self, gp_ods_for_patient):
...
+        )
+
+        self.db_service.update_item(
+            table_name=self.session_table_name,
+            key_pair={"NDRSessionId": ndr_session_id},
+            updated_fields=updated_fields,
+            condition_expression=(
+                f"attribute_not_exists({self.permitted_field})"
+                if not allowed_nhs_numbers
+                else None
```
steph-torres-nhs, line 133: is this needed? safety net?

lambdas/tests/unit/services/test_authoriser_service.py
```diff
@@ -90,18 +92,49 @@ def test_deny_access_policy_returns_true_for_gp_clinical_on_paths(
...
+        test_path, RepositoryRole.GP_CLINICAL.value, "900000001"
+    )
+    assert actual == expected
+    mock_auth_service.allowed_nhs_numbers = []
+
+
+@pytest.mark.parametrize(
+    "test_path", ["/DocumentManifest", "/DocumentDelete", "/DocumentReference", "Any"]
+)
+def test_deny_access_policy_returns_true_for_nhs_number_not_in_allowed(
```
steph-torres-nhs, line 106: do we want a test for returning false nhs_number_in_allowed?

lambdas/services/authoriser_service.py
```diff
@@ -15,12 +15,13 @@
 
 
 class AuthoriserService:
-    def __init__(
-        self,
-    ):
+    def __init__(self):
         self.redact_session_id = ""
+        self.allowed_nhs_numbers = []
```
steph-torres-nhs, line 20: would using a set here help us?

lambdas/services/authoriser_service.py
abbas-khan10: Good spot, however "AllowedNHSNumbers" will be used to turn create a list on line 96, so by defaulting here we're reducing a check later to set `self.allowed_nhs_numbers` to an empty list

lambdas/services/search_patient_details_service.py
abbas-khan10: Yep, updating an item with the conditional attriubute depends on whether "AllowedNHSNumbers" exists, which won't exist on the users first search, but will in subsequent searches. So having a ternary check here is deciding whether to add the conditional check if the user has `allowed_nhs_numbers` marked against that session. If you try to update something with a field that doesn't exist, dynamo will error

lambdas/services/authoriser_service.py
abbas-khan10: For uniqueness, definitely. However the logic (Line 115 in `lambdas/services/search_patient_details_service.py`)that decides if we add to the list will remain the same, we are checking to see if the value exists in a list to decide whether or not we write to dynamo. So I think the benefit of using a set here is lost

steph-torres-nhs (approved):

AndyFlintNHS (approved):

lambdas/tests/unit/services/test_authoriser_service.py
NogaNHS: Done

PR #540, opened 2025-02-10
[PRMP-1522] Auth changes to upload and manifest journey

NogaNHS:


lambdas/handlers/upload_confirm_result_handler.py
```diff
@@ -50,9 +52,14 @@ def lambda_handler(event, context):
     if not upload_lambda_enabled_flag_object[upload_flag_name]:
         logger.info("Feature flag not enabled, event will not be processed")
         raise FeatureFlagsException(500, LambdaError.FeatureFlagDisabled)
-    nhs_number, documents = processing_event_details(event)
-    request_context.patient_nhs_no = nhs_number
-    upload_confirm_result_service = UploadConfirmResultService(nhs_number)
+    nhs_number_query_string = event["queryStringParameters"]["patientId"]
```
abbas-khan10, line 55: Any reason we're calling this `nhs_number_query_string` and everywhere else it's just `nhs_number`?

lambdas/services/search_patient_details_service.py
```diff
@@ -35,6 +44,9 @@ def handle_search_patient_request(self, nhs_number):
             logger.audit_splunk_info(
                 "Searched for patient details", {"Result": "Patient found"}
             )
+
+            self.update_auth_session_with_permitted_search(nhs_number=nhs_number)
```
MohammadIqbalAD-NHS, line 48: Could this be modified to:
```suggestion
            self.update_auth_session_with_permitted_search(nhs_number)
```
?

lambdas/services/update_upload_state_service.py
```diff
@@ -46,15 +47,21 @@ def handle_update_state(self, event_body: dict):
...
         updated_fields = self.format_update({Fields.UPLOADING.value: uploaded})
         table = doc_type.get_dynamodb_table_name()
         try:
             self.dynamo_service.update_item(
                 table_name=table,
-                key=doc_ref,
+                key_pair={DocumentReferenceMetadataFields.ID.value: doc_ref},
                 updated_fields=updated_fields,
+                condition_expression="begins_with(NhsNumber, :v_nhs_number)",
+                expression_attribute_values={":v_nhs_number": nhs_number},
```
MohammadIqbalAD-NHS, line 64: Could you rename by specifying what the 'v' in `v_nhs_number` stands for, as it was not apparent to me.

NogaNHS: Taken from this example - https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#Expressions.ConditionExpressions.ConditionalExamples

MohammadIqbalAD-NHS: Given that this is AWS' preferred practice, i'm fine with it.

MohammadIqbalAD-NHS (approved):

abbas-khan10 (approved):

lambdas/services/search_patient_details_service.py
abbas-khan10: It's a valid comment, however I think for clarity and readability we want to enforce function calls with the same pattern `function(function_param_name=parameter_value)`

PR #542, opened 2025-02-11
[PRMP-1523] Change to role permissions 

NogaNHS:


abbas-khan10 (approved):

steph-torres-nhs (approved):

AndyFlintNHS:
Happy with this as it is in line with the documentation here - https://nhse-dsic.atlassian.net/wiki/spaces/TW/pages/12535824385/Role-based+Access+Requirements

It's worth mentioning that the ticket implies that Health Records Manager (GP_ADMIN role) shouldn't have access to uploading via the API, but we've chosen the documentation as the source of truth.

PR #544, opened 2025-02-12
[PRMP-1535] Add object-src to CSP

steph-torres-nhs:


MohammadIqbalAD-NHS (approved):

chrisbloe-nhse (approved):
Happy with the `object-src` being added.

PR #545, opened 2025-02-14
[PRMP-1427] Report by ODS Code - lambda

NogaNHS:


lambdas/tests/unit/helpers/data/dynamo/dynamo_stream.py
```diff
@@ -1,50 +1,50 @@
...
-    ]
-}
-
-MOCK_OLD_IMAGE_EVENT = MOCK_DYNAMO_STREAM_EVENT["Records"][0]["dynamodb"]["OldImage"]
-
-MOCK_OLD_IMAGE_MODEL = DocumentReference.model_validate(
-    parse_dynamo_record(MOCK_OLD_IMAGE_EVENT)
-)
+from models.document_reference import DocumentReference
+from tests.unit.conftest import (
```
abbas-khan10, line 2: What's happening with these entire file changes? Formatting maybe?

lambdas/utils/decorators/validate_dynamo_stream_event.py
```diff
@@ -1,47 +1,47 @@
...
-                return ApiGatewayResponse(
-                    400,
-                    LambdaError.DynamoInvalidStreamEvent.create_error_body(),
-                    event.get("httpMethod", "GET"),
-                ).create_api_gateway_response()
-
-        return lambda_func(event, context)
-
-    return interceptor
+from typing import Callable
```
abbas-khan10, line 1: This file too

lambdas/utils/dynamo_query_filter_builder.py
```diff
@@ -24,6 +24,8 @@ def add_condition(
                 condition = Attr(attribute).lt(filter_value)
             case AttributeOperator.LESS_THAN_OR_EQUAL:
                 condition = Attr(attribute).lte(filter_value)
+            case AttributeOperator.IN:
+                condition = Attr(attribute).is_in(filter_value)
```
abbas-khan10, line 28: Please could we add a test case for this?

Also I know it's not part of your ticket but I always wanted to refactor this statement as it became much larger than originally anticipated. Because we have the enum for the attribute which match the method calls, a better solution would be to replace the entire function with this block. All unit tests for the query builder should still pass without needed to do anything else, and it's much cleaner:
```

def add_condition(
        self, attribute: str, attr_operator: AttributeOperator, filter_value
    ):
        try:
            condition = getattr(Attr(attribute), attr_operator.value)(filter_value)
        except AttributeError:
            raise DynamoServiceException(
                f"Unsupported attribute filter operator: {attr_operator}"
            )
        self.filter_conditions.append(condition)
        return self
```

lambdas/tests/unit/helpers/data/dynamo/dynamo_stream.py
NogaNHS: line ending change

lambdas/utils/dynamo_query_filter_builder.py
NogaNHS: Done

lambdas/services/ods_report_service.py
```diff
@@ -0,0 +1,224 @@
...
+            + ods_code
+            + f"_{len(nhs_numbers)}_"
+            + formatted_time
+            + "."
+            + file_type_output
+        )
+        temp_file_path = os.path.join(self.temp_output_dir, file_name)
+        match file_type_output:
+            case FileType.CSV:
+                self.create_report_csv(temp_file_path, nhs_numbers, ods_code)
```
AndyFlintNHS, line 151: Very nitpicky but can we rename this to create_csv_report to be named in line with the other reports?

AndyFlintNHS (changes requested):

lambdas/services/authoriser_service.py
```diff
@@ -101,6 +101,8 @@ def deny_access_policy(self, path, user_role, nhs_number: str = None):
             case "/SearchPatient":
                 deny_resource = False
 
+            case "/OdsReport":
+                deny_resource = deny_access_to_clinical_role
```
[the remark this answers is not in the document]
abbas-khan10: Valid spot, I don't believe patient search restriction is required here as this is not part of the search journey

lambdas/enums/dynamo_filter.py
```diff
@@ -8,6 +8,7 @@ class AttributeOperator(Enum):
     GREATER_OR_EQUAL = "gte"
     LESS_THAN = "lt"
     LESS_THAN_OR_EQUAL = "lte"
+    IN = "is_in"
```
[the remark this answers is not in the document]
abbas-khan10: The boto3 docs does show the implementation for `IN` is `is_in` https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/dynamodb/conditions.html

lambdas/handlers/get_report_by_ods_handler.py
```diff
@@ -0,0 +1,81 @@
...
+        "org_ods_code"
+    )
+    if not ods_code:
+        raise OdsErrorException("No ODS code provided")
+    file_type = event.get("queryStringParameters", {}).get("OutputFileFormat")
+    if file_type not in FileType.__members__.values():
+        file_type = FileType.CSV
+    service = OdsReportService()
+    logger.info(f"Starting process for ods code: {ods_code}")
+    pre_sign_url = service.get_nhs_numbers_by_ods(
```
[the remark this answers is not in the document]
abbas-khan10: Will change this

lambdas/scripts/create_report_by_ods.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+from enums.file_type import FileType
+from services.ods_report_service import OdsReportService
+
+if __name__ == "__main__":
+    ods_code = "_"
```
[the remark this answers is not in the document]
abbas-khan10: This was a manual script to be run, I'm unsure what it's purpose serves outside of testing so wonder if it is worth even keeping in

lambdas/services/ods_report_service.py
```diff
@@ -0,0 +1,224 @@
...
+                exclusive_start_key=response["LastEvaluatedKey"],
+                table_name=self.table_name,
+                filter_expression=ods_filter_expression,
+            )
+            results.extend(response.get("Items", []))
+        if not results:
+            logger.info("No records found for ODS code {}".format(ods_code))
+            raise OdsReportException(404, LambdaError.NoDataFound)
+
+        return results
```
steph-torres-nhs, line 87: do we want to use dynamo service scan whole table here?

lambdas/services/ods_report_service.py
```diff
@@ -0,0 +1,224 @@
...
+        ods_code: str,
+        is_pre_signed_needed: bool = False,
+        is_upload_to_s3_needed: bool = False,
+        file_type_output: FileType = FileType.CSV,
+    ):
+        results = self.scan_table_with_filter(ods_code)
+        nhs_numbers = {
+            item.get(DocumentReferenceMetadataFields.NHS_NUMBER.value)
+            for item in results
+            if item.get(DocumentReferenceMetadataFields.NHS_NUMBER.value)
```
steph-torres-nhs, line 44: Is there any case where we don't get an NHS number here?

lambdas/services/ods_report_service.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def get_nhs_numbers_by_ods(
+        self,
+        ods_code: str,
+        is_pre_signed_needed: bool = False,
+        is_upload_to_s3_needed: bool = False,
+        file_type_output: FileType = FileType.CSV,
+    ):
+        results = self.scan_table_with_filter(ods_code)
+        nhs_numbers = {
```
steph-torres-nhs, line 41: Do we want to be explicit that this is a set?

lambdas/handlers/get_report_by_ods_handler.py
```diff
@@ -0,0 +1,82 @@
...
+from utils.lambda_response import ApiGatewayResponse
+from utils.request_context import request_context
+
+logger = LoggingService(__name__)
+
+
+@ensure_environment_variables(
+    names=[
+        "LLOYD_GEORGE_DYNAMODB_NAME",
+    ]
```
steph-torres-nhs, line 24: does this need "STATISTICAL_REPORTS_BUCKET"  too?

lambdas/services/ods_report_service.py
abbas-khan10: I believe we are not adding a GSI to improve the speed of the query here until we get some initial feedback from the performance testing

lambdas/services/ods_report_service.py
abbas-khan10: There is no need, the syntax should be enough however I have added typing to the method params to be clear

lambdas/services/ods_report_service.py
steph-torres-nhs: can we remove the if statement then?

abbas-khan10: No because then it would add a `None` object to the set

AndyFlintNHS (approved):

PR #547, opened 2025-02-19
[PRMP-1541] Content-Security-Policy header -> Add object-src <CloudFront domain>

AndyFlintNHS:


chrisbloe-nhse (approved):
It would be preferable to have two separate `sed` commands, but ok to run as one.

abbas-khan10 (approved):

PR #554, opened 2025-02-24
[PRMP-1512] - Add message to sqs-stitching queue during bulk upload

MohammadIqbalAD-NHS:


lambdas/models/pdf_stitcher_sqs_message.py
```diff
@@ -0,0 +1,12 @@
+from enums.snomed_codes import SnomedCode, SnomedCodes
+from pydantic import AliasGenerator, BaseModel, ConfigDict
+from pydantic.alias_generators import to_camel
+
+
+class PdfStitcherSqsMessage(BaseModel):
+    model_config = ConfigDict(
+        alias_generator=AliasGenerator(serialization_alias=to_camel),
```
steph-torres-nhs, line 8: does this need `use_enum_values=True` like NrlSqsMessage model?

lambdas/services/bulk_upload_service.py
```diff
@@ -251,9 +249,7 @@ def handle_sqs_message(self, message: dict):
         )
 
         try:
-            last_document_processed = self.create_lg_records_and_copy_files(
```
steph-torres-nhs, line 254: does the return value from this function need removing too?

lambdas/repositories/bulk_upload/bulk_upload_sqs_repository.py
```diff
@@ -42,11 +42,10 @@ def put_sqs_message_back_to_queue(self, sqs_message: dict):
...
 
-    def send_message_to_nrl_fifo(
-        self, queue_url: str, message: NrlSqsMessage, group_id: str
+    def send_message_to_pdf_stitcher_queue(
+        self, queue_url: str, message: PdfStitcherSqsMessage
     ):
-        self.sqs_repository.send_message_fifo(
+        self.sqs_repository.send_message_standard(
             queue_url=queue_url,
             message_body=message.model_dump_json(),
```
abbas-khan10, line 50: I believe this should be model_dump_json(by_alias=True) so we are outputting camel case (keeping the message format as json)

lambdas/models/pdf_stitcher_sqs_message.py
```diff
@@ -0,0 +1,12 @@
...
+    )
+
+    nhs_number: str
+    snomed_code: SnomedCode = SnomedCodes.LLOYD_GEORGE.value
```
abbas-khan10, line 12: Sorry I know I suggested this change but after chatting with Kris it looks like this is referred to as `snomed_code_doc_type` everywhere else so could this be changed back please

lambdas/services/bulk_upload_service.py
MohammadIqbalAD-NHS: Yes! Has been removed.

lambdas/models/pdf_stitcher_sqs_message.py
MohammadIqbalAD-NHS: I have reverted it.

steph-torres-nhs (approved):

PR #558, opened 2025-02-28
[PRMP-1579] Add DLQ to MNS

steph-torres-nhs:


lambdas/services/process_mns_message_service.py
```diff
@@ -34,9 +34,12 @@ def handle_mns_notification(self, message: MNSSQSMessage):
...
                     self.handle_death_notification(message)
 
-        except PdsErrorException:
+        except PdsErrorException as e:
             logger.info("An error occurred when calling PDS")
-            self.send_message_back_to_queue(message)
+            logger.info(
+                f"Unable to process message: {message.id}, of type: {message.type}"
+            )
+            logger.info(f"{e}")
```
NogaNHS, line 42: Are we catching this in the handler?
I think we need to remove the decorator from the handler as well

MohammadIqbalAD-NHS (approved):

lambdas/handlers/mns_notification_handler.py
```diff
@@ -38,13 +34,18 @@ def lambda_handler(event, context):
...
             if mns_message.type in MNSNotificationTypes.__members__.values():
                 notification_service.handle_mns_notification(mns_message)
 
         except ValidationError as error:
             logger.error("Malformed MNS notification message")
             logger.error(error)
+            raise error
         except Exception as error:
             logger.error(f"Error processing SQS message: {error}.")
+            raise error
```
AndyFlintNHS, line 50: Can we modify the logging here so that they mention that the message will be placed on the DLQ?

lambdas/handlers/mns_notification_handler.py
```diff
@@ -23,8 +21,6 @@
         "MNS_NOTIFICATION_QUEUE_URL",
     ]
 )
-@override_error_check
```
AndyFlintNHS, line 26: This doesn't need to be removed.

lambdas/handlers/mns_notification_handler.py
AndyFlintNHS: Discussed with Steph - this isn't a good idea as the messages will be placed on the DLQ by SQS after 3 attempts and logging this would be misleading on the first 2 failures.

NogaNHS (approved):

abbas-khan10 (approved):

PR #564, opened 2025-03-05
[PRMP-1585] Allow access to deceased patients

NogaNHS:


lambdas/services/manage_user_session_access.py
```diff
@@ -0,0 +1,105 @@
...
+
+        self.session_table_name = os.getenv("AUTH_SESSION_TABLE_NAME")
+        self.permitted_field = "AllowedNHSNumbers"
+        self.deceased_field = "DeceasedNHSNumbers"
+
+    def find_login_session(self, ndr_session_id: str):
+        logger.info(
+            f"Retrieving session for session ID ending in: f{ndr_session_id[:-4]}"
+        )
+        session_table_name = os.environ["AUTH_SESSION_TABLE_NAME"]
```
abbas-khan10, line 24: duplicate `os.environ` above?

lambdas/services/manage_user_session_access.py
```diff
@@ -0,0 +1,105 @@
...
+        query_response = self.db_service.query_all_fields(
+            table_name=session_table_name,
+            search_key="NDRSessionId",
+            search_condition=ndr_session_id,
+        )
+
+        try:
+            current_session = query_response["Items"][0]
```
abbas-khan10, line 32: I wonder if we need a pydantic model to hold and validate our session table data to improve error handling and the passing around of these object for any future changes regading the auth?

abbas-khan10, line 32: Also will element 0 always return the latest?

lambdas/services/manage_user_session_access.py
```diff
@@ -0,0 +1,105 @@
...
+
+        self.db_service.update_item(
+            table_name=self.session_table_name,
+            key_pair={"NDRSessionId": ndr_session_id},
+            updated_fields=updated_fields,
+            condition_expression=(
+                f"attribute_not_exists({field_name})"
+                if not existing_nhs_numbers
+                else None
+            ),
```
AndyFlintNHS, line 91: I know this was present in the previous code and has been copied over, however, is this redundant? Here is DynamoDB's documentation:
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html

The way that we're reading this part: `The UpdateItem operation updates an existing item, or adds a new item to the table if it does not already exist.` is that _'If the field doesn't exist, an update_item should create this'_

AndyFlintNHS (changes requested):
Note - I have not finished my review, but other priorities have come up, so these are my comments

lambdas/services/manage_user_session_access.py
```diff
@@ -0,0 +1,105 @@
...
+        allowed_nhs_numbers = current_session.get(self.permitted_field, False)
+        deceased_nhs_numbers = current_session.get(self.deceased_field, False)
+        logger.info("Searching Auth session table for existing NHS number")
+        if nhs_number in allowed_nhs_numbers:
+            logger.info(
+                "Permitted search, NHS number already exists in allowed NHS numbers"
+            )
+            return
+
+        if deceased:
```
steph-torres-nhs, line 54: do we need a case for deceased and Role == to PCSE to add nhs number to both allowed and deceased fields.

lambdas/services/manage_user_session_access.py
NogaNHS: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.CoreComponents.html#HowItWorks.CoreComponents.PrimaryKey

lambdas/services/manage_user_session_access.py
NogaNHS: @abbas-khan10 I remember we talked about it last time, do we need that?

lambdas/services/manage_user_session_access.py
NogaNHS: I think possibly out of the scope of this ticket?

lambdas/services/manage_user_session_access.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+
+    def update_auth_session_with_permitted_search(
+        self, user_role: RepositoryRole, nhs_number: str, deceased: bool
+    ):
+        ndr_session_id = request_context.authorization.get("ndr_session_id")
+        current_session = self.find_login_session(ndr_session_id)
+
+        allowed_nhs_numbers = current_session.get(self.permitted_field, "")
+        deceased_nhs_numbers = current_session.get(self.deceased_field, "")
```
NogaNHS, line 46: split those into lists

lambdas/services/manage_user_session_access.py
NogaNHS: Done

AndyFlintNHS (approved):

PR #565, opened 2025-03-05
[PRMP-1511] PDF Stitcher - Generate single PDF

abbas-khan10:


lambdas/handlers/pdf_stitching_handler.py
```diff
@@ -0,0 +1,62 @@
...
+@handle_lambda_exceptions
+def lambda_handler(event, context):
+    request_context.app_interaction = LoggingAppInteraction.STITCH_RECORD.value
+
+    logger.info("Received PDF Stitching SQS message event")
+    event_message_records = event.get("Records")
+    pdf_stitching_service = PdfStitchingService()
+
+    for message in event_message_records:
+        try:
```
NogaNHS, line 43: Maybe reset request_context.patient_nhs_no here. `request_context.patient_nhs_no = ""`
Otherwise if ValidationError is caught it would log with the previous number

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,199 @@
...
+            )
+        except ClientError as e:
+            logger.error(f"Failed to create stitching reference: {e}")
+            raise PdfStitchingException(400, LambdaError.StitchError)
+
+    def publish_nrl_message(self):
+        document_api_endpoint = (
+            os.environ.get("APIM_API_URL", "")
+            + "/DocumentReference/"
+            + SnomedCodes.LLOYD_GEORGE.value.code
```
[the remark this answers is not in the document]
abbas-khan10: It is initally, we're only stitching bulk upload at the moment I brought this up with regards to create doc ref but it was out of scope and needs further discussion how we will expand this for a wider use case

steph-torres-nhs (approved):

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,204 @@
...
+    def migrate_multipart_references(
+        self, multipart_references: list[DocumentReference]
+    ):
+        logger.info("Migrating multipart references")
+        try:
+            for reference in multipart_references:
+                self.dynamo_service.create_item(
+                    table_name=self.unstitched_lloyd_george_table_name,
+                    item=reference.model_dump_dynamo(),
+                )
```
NogaNHS, line 132: Can you use `delete_document_references` from document_service here?

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,204 @@
...
+        except ClientError as e:
+            logger.error(f"Failed to migrate multipart references: {e}")
+            raise PdfStitchingException(400, LambdaError.MultipartError)
+
+        try:
+            for reference in multipart_references:
+                self.dynamo_service.delete_item(
+                    table_name=self.lloyd_george_table_name,
+                    key={DocumentReferenceMetadataFields.ID.value: reference.id},
+                )
```
NogaNHS, line 142: Can you use `delete_document_references` from document_service here?

lambdas/services/pdf_stitching_service.py
NogaNHS, line 132: Is it worthwhile to create a method in the document service for creating items?

lambdas/services/pdf_stitching_service.py
abbas-khan10: We are not soft deleting here, we are removing the dynamo entry entirely and moving the references to the unstitched table. The `delete_document_references` would keep them in the reference table and just give them a TTL (which isn't what we want here)

chrisbloe-nhse (dismissed):
`CurrentGpOds` shouldn't be copied to the new unstitched table...

AndyFlintNHS (approved):

lambdas/models/document_reference.py
```diff
@@ -88,6 +88,27 @@ def last_updated_within_three_minutes(self) -> bool:
...
+            DocumentReferenceMetadataFields.UPLOADED.value: self.uploaded,
+            DocumentReferenceMetadataFields.UPLOADING.value: self.uploading,
+            DocumentReferenceMetadataFields.LAST_UPDATED.value: self.last_updated,
+        }
+
+        if self.ttl is not None:
+            dynamo_model[DocumentReferenceMetadataFields.TTL.value] = self.ttl
+
+        return dynamo_model
+
```
NogaNHS, line 111: Can we use model_dump_json? with exclude None?

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+                doc_type=SupportedDocumentTypes.LG,
+                query_filter=UploadCompleted,
+            )
+        )
+        if len(document_references) == 0 or any(
+            "1of1" in document_reference.file_name
+            for document_reference in document_references
+        ):
+            logger.info("No usable files found for stitching")
+            return
```
NogaNHS, line 56: Can we filter by this? so we don't get it back?

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+        )
+
+        self.upload_stitched_file(stitching_data_stream=stitching_data_stream)
+        self.migrate_multipart_references(multipart_references=document_references)
+        self.write_stitching_reference()
+        self.publish_nrl_message()
+
+    def create_stitched_reference(self, document_reference: DocumentReference):
+        date_now = datetime.now(timezone.utc)
+        reference_id = create_reference_id()
```
NogaNHS, line 74: Could we let pydantic create the id?

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+        stripped_filename = re.sub(r"^\d+of\d+_", "", document_reference.file_name)
+
+        self.stitched_reference = DocumentReference(
+            id=reference_id,
+            content_type=document_reference.content_type,
+            created=date_now.strftime(DATE_FORMAT),
+            current_gp_ods=document_reference.current_gp_ods,
+            deleted="",
```
NogaNHS, line 82: Can we add this as a default to the model?

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+            file_location=f"s3://{self.lloyd_george_bucket_name}/{document_reference.nhs_number}/{reference_id}",
```
NogaNHS, line 83: Can be created by the model using computed field?

lambdas/services/pdf_stitching_service.py
abbas-khan10: Good idea, yes we should be able to will add this

lambdas/models/document_reference.py
abbas-khan10: Unfortunately I don't think so, as the model alias gen is set to output camel case, and without would result in snake_case. This function is specific to a newly created dynamo item, but will rename the function to be a bit clearer

lambdas/services/pdf_stitching_service.py
abbas-khan10: I think because this model now handles both reading existing dynamo items and writing new items, pre-populating this with an ID might not be a good idea as there's a risk of defaulting to a new ID instead of the existing ID on model validation. Better to be specific in the creation of a new ID for new items outside the model for now, until these models are revisited and cleaned up?

lambdas/services/pdf_stitching_service.py
abbas-khan10: Same reason above for the ID

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+            raise PdfStitchingException(400, LambdaError.StitchError)
+
+    def migrate_multipart_references(
+        self, multipart_references: list[DocumentReference]
+    ):
+        logger.info("Migrating multipart references")
+        try:
+            for reference in multipart_references:
+                dynamo_item = reference.model_dump_dynamo()
+                dynamo_item.pop(DocumentReferenceMetadataFields.CURRENT_GP_ODS.value)
```
NogaNHS, line 130: model_dump_json(exclude=DocumentReferenceMetadataFields.CURRENT_GP_ODS.value)

lambdas/tests/unit/services/test_pdf_stitching_service.py
```diff
@@ -0,0 +1,425 @@
...
+    with pytest.raises(PdfStitchingException) as e:
+        mock_service.publish_nrl_message()
+
+    assert caplog.records[-1].msg == expected_err_msg
+    assert caplog.records[-1].levelname == "ERROR"
+    assert e.value.error is LambdaError.StitchError
+
+
+def test_publish_nrl_message_handles_error():
+    pass
```
NogaNHS, line 401: Missing test?

lambdas/tests/unit/services/test_pdf_stitching_service.py
```diff
@@ -0,0 +1,425 @@
...
+
+@pytest.fixture
+def mock_publish_nrl_message(mocker, mock_service):
+    return mocker.patch.object(mock_service, "publish_nrl_message")
+
+
+@pytest.fixture
+def mock_download_fileobj():
+    def _mock_download_fileobj(
+        s3_object_data: dict[str, BytesIO], Bucket: str, Key: str, Fileobj: BytesIO
```
NogaNHS, line 86: Change params to snake_case?

lambdas/tests/unit/services/test_pdf_stitching_service.py
abbas-khan10: Good spot, obviously didn't think it was necessary...

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+            last_updated=int(date_now.timestamp()),
+        )
+
+    def process_stitching(self, s3_object_keys: list[str]) -> BytesIO:
+        pdf_writer = PdfWriter()
+
+        for key in s3_object_keys:
+            data_stream = BytesIO()
+            self.s3_service.client.download_fileobj(
+                Bucket=self.lloyd_george_bucket_name, Key=key, Fileobj=data_stream
```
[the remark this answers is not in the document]
abbas-khan10: They don't follow snake case it looks like https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/download_fileobj.html

lambdas/services/pdf_stitching_service.py
NogaNHS: I think pydantic should be able to handle both existing id and creating new ones, I'll look in the docs.

lambdas/models/document_reference.py
NogaNHS: `model_dump_json(by_alias=True, exclude_none=True)`?

lambdas/services/pdf_stitching_service.py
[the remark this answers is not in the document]
abbas-khan10: I don't think we need another method, we're just using the method boto3 provides, might be worth wrapping in a try catch in case of failure to log our the response but don't think we need anything further

lambdas/handlers/pdf_stitching_handler.py
```diff
@@ -0,0 +1,63 @@
...
+            request_context.patient_nhs_no = ""
+            message_body = json.loads(message.get("body", ""))
+            stitching_message = PdfStitchingSqsMessage.model_validate(message_body)
```
NogaNHS, line 46: Where are you checking the SNOMED in the message is LG?

lambdas/handlers/pdf_stitching_handler.py
```diff
@@ -0,0 +1,63 @@
...
+        "LLOYD_GEORGE_DYNAMODB_NAME",
+        "UNSTITCHED_LLOYD_GEORGE_DYNAMODB_NAME",
+        "PDF_STITCHING_SQS_URL",
+        "APIM_API_URL",
+        "NRL_SQS_URL",
+    ]
+)
+@override_error_check
+@validate_sqs_event
+@handle_lambda_exceptions
```
NogaNHS, line 34: This would prevent failed messages from getting to the DLQ as the lambda would handle the exceptions.

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+            )
+        except ClientError as e:
+            logger.error(f"Failed to create stitching reference: {e}")
+            raise PdfStitchingException(400, LambdaError.StitchError)
+
+    def publish_nrl_message(self):
+        document_api_endpoint = (
+            os.environ.get("APIM_API_URL", "")
+            + "/DocumentReference/"
+            + SnomedCodes.LLOYD_GEORGE.value.code
```
NogaNHS, line 163: Can we pass in the snomed from the original message? So we're making this more dynamic to future use cases?

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,206 @@
...
+        )
+
+        try:
+            self.sqs_service.send_message_fifo(
+                queue_url=os.environ.get("NRL_SQS_URL"),
+                message_body=nrl_sqs_message.model_dump_json(),
+                group_id=f"nrl_sqs_{uuid.uuid4()}",
+            )
+        except ClientError as e:
+            logger.error(f"Failed to publish NRL message onto SQS: {e}")
```
NogaNHS, line 184: The ticket mention sending messages to the DLQ upon failure. But if the process fail here there is no records in the original table anymore, so a re-drive from the DLQ would do nothing to complete the process

lambdas/handlers/pdf_stitching_handler.py
abbas-khan10: Good spot this was missed from the last PR which introduced the model, the snomed code was set as a default to LG. I've removed it so it validates against the model now

lambdas/models/document_reference.py
```diff
@@ -88,6 +98,27 @@ def last_updated_within_three_minutes(self) -> bool:
...
+            DocumentReferenceMetadataFields.CURRENT_GP_ODS.value: self.current_gp_ods,
+            DocumentReferenceMetadataFields.UPLOADED.value: self.uploaded,
+            DocumentReferenceMetadataFields.UPLOADING.value: self.uploading,
+            DocumentReferenceMetadataFields.LAST_UPDATED.value: self.last_updated,
+        }
+
+        if self.ttl is not None:
+            dynamo_model[DocumentReferenceMetadataFields.TTL.value] = self.ttl
+
+        return dynamo_model
```
NogaNHS, line 120: Can be removed with the config changes?

lambdas/services/document_deletion_service.py
```diff
@@ -88,7 +89,10 @@ def delete_documents_references_in_stitch_table(self, nhs_number: str):
             self.document_service.dynamo_service.update_item(
                 table_name=self.stitch_service.stitch_trace_table,
                 key_pair={DocumentReferenceMetadataFields.ID.value: record.id},
-                updated_fields=record.model_dump(by_alias=True, include={"deleted"}),
+                updated_fields=record.model_dump(
+                    by_alias=True,
+                    include={underscore(DocumentReferenceMetadataFields.DELETED.value)},
```
NogaNHS, line 94: Wouldn't underscore exclude delete? Why do we need it here?

abbas-khan10: No it will just turn the enum value into snakecase so we're not hard coding strings in our queries. Reduces risk of error and enforces some consistency

lambdas/models/sqs/pdf_stitching_sqs_message.py
```diff
@@ -0,0 +1,7 @@
+from enums.snomed_codes import SnomedCode
+from pydantic import BaseModel
+
+
+class PdfStitchingSqsMessage(BaseModel):
+    nhs_number: str
+    snomed_code_doc_type: SnomedCode
```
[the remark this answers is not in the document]
abbas-khan10: I didn't think we should be defaulting to anything, we shouldn't assume the type of stitch. The message should contain the correct doc type

lambdas/services/pdf_stitching_service.py
```diff
@@ -0,0 +1,310 @@
...
+
+        return document_references
+
+    def process_message(self, stitching_message: PdfStitchingSqsMessage):
+        doc_type = (
+            SupportedDocumentTypes.LG
+            if stitching_message.snomed_code_doc_type.code
+            == SnomedCodes.LLOYD_GEORGE.value.code
+            else SupportedDocumentTypes.ARF
+        )
```
NogaNHS, line 71: Do we stitch ARF docs?

NogaNHS (approved):

PR #568, opened 2025-03-07
[PRMP-5115] Rename smoke test file

bethany-kish-nhs:
Rename smoke test file so it gets picked up by cypress


MohammadIqbalAD-NHS (approved):

NogaNHS (approved):

chrisbloe (changes requested):
The branch name is incorrect!

PR #571, opened 2025-03-11
[PRMP-1601] New AccessAudit lambda

NogaNHS:


lambdas/handlers/access_audit_handler.py
```diff
@@ -0,0 +1,31 @@
...
+
+@set_request_context_for_logging
+@validate_patient_id
+@override_error_check
+@ensure_environment_variables(
+    names=["AUTH_SESSION_TABLE_NAME", "ACCESS_AUDIT_TABLE_NAME"]
+)
+@handle_lambda_exceptions
+def lambda_handler(event, context):
+    nhs_number = event.get("queryStringParameters", {}).get("patientId")
```
abbas-khan10, line 23: We have the validate id decorator above, do we need to add this extra logic to populate an empty dict and do the conditional check on line 26?

lambdas/enums/deceased_access_reason.py
```diff
@@ -0,0 +1,10 @@
+from utils.multi_value_enum import MultiValueEnum
+
+
+class DeceasedAccessReason(MultiValueEnum):
+    MEDICAL = "01", "Coroner or medical examiner request"
+    LEGAL = "02", "Legal or insurance request"
+    PERSONAL = "03", "Family member or personal representation request"
+    INTERNAL = "04", "Internal NHS request"
+    MANAGE = "05", "Manage this record at the end of its retention period"
+    OTHER = "06", "Another reason"
```
adamwhitingnhs, line 10: I think other is supposed to be code "99"

lambdas/handlers/access_audit_handler.py
```diff
@@ -0,0 +1,31 @@
...
+@set_request_context_for_logging
+@validate_patient_id
+@override_error_check
+@ensure_environment_variables(
+    names=["AUTH_SESSION_TABLE_NAME", "ACCESS_AUDIT_TABLE_NAME"]
+)
+@handle_lambda_exceptions
+def lambda_handler(event, context):
+    nhs_number = event.get("queryStringParameters").get("patientId")
+    request_type = event.get("queryStringParameters", {}).get("accessAuditType")
```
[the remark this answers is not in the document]
NogaNHS: No, because we have the decorator

steph-torres-nhs (approved):

PR #578, opened 2025-03-18
[PRMP-1580] NHSOAuthTokenGenerator lambda

AndyFlintNHS:


NogaNHS:
Do we also want to remove the current writing to the same ssm parameter in other lambdas?

lambdas/services/base/nhs_oauth_service.py
```diff
@@ -32,6 +33,9 @@ def get_active_access_token(self):
         return access_token
 
     def get_new_access_token(self):
+        return self.get_nhs_oauth_response()["access_token"]
```
NogaNHS, line 36: Could get KeyError here. maybe use `.get` with default instead

lambdas/tests/unit/handlers/test_nhs_oauth_token_generator_handler.py
```diff
@@ -0,0 +1,70 @@
...
+    mock = mocker.Mock()
+    mocker.patch(
+        "handlers.nhs_oauth_token_generator_handler.NhsOauthService", return_value=mock
+    )
+    return mock
+
+
+@pytest.fixture
+def mock_logger(mocker):
+    return mocker.patch("handlers.nhs_oauth_token_generator_handler.logger")
```
NogaNHS, line 22: Don't need to mock logger, we have `logger_mocker` in conftest that should cover that.

lambdas/handlers/nhs_oauth_token_generator_handler.py
```diff
@@ -0,0 +1,42 @@
...
+from utils.decorators.override_error_check import override_error_check
+from utils.decorators.set_audit_arg import set_request_context_for_logging
+from utils.exceptions import OAuthErrorException
+
+logger = LoggingService(__name__)
+
+
+@set_request_context_for_logging
+@override_error_check
+@handle_lambda_exceptions
```
NogaNHS, line 17: Do we what this here? We should make sure we get an alarm if this lambda failed

lambdas/tests/unit/handlers/test_nhs_oauth_token_generator_handler.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_lambda_handler_fails_after_max_retries(
+    mock_auth_service, mock_logger, mock_sleep, event, context
+):
+    mock_auth_service.get_nhs_oauth_response.side_effect = OAuthErrorException(
+        "Error creating OAuth access token"
+    )
+
+    lambda_handler(
```
NogaNHS, line 62: You can still assert the error the lambda is returning

lambdas/tests/unit/services/base/test_nhs_oauth_service.py
NogaNHS, line 1: Maybe change the tests to test 'get_nhs_oauth_response' instead of 'get_new_access_token', and add one new test for  'get_new_access_token'. At the moment they are testing both.

lambdas/tests/unit/handlers/test_nhs_oauth_token_generator_handler.py
abbas-khan10: If you do want to test for specific log messages you can use `caplog` as used in other tests

lambdas/tests/unit/handlers/test_nhs_oauth_token_generator_handler.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@pytest.fixture
+def mock_logger(mocker):
+    return mocker.patch("handlers.nhs_oauth_token_generator_handler.logger")
+
+
+@pytest.fixture
+def mock_sleep(monkeypatch):
+    monkeypatch.setattr(time, "sleep", lambda _: None)
```
abbas-khan10, line 27: What is the purpose of this fixture?

lambdas/tests/unit/handlers/test_nhs_oauth_token_generator_handler.py
AndyFlintNHS: Opted to remove as we're now testing for the error being raised instead

lambdas/handlers/nhs_oauth_token_generator_handler.py
AndyFlintNHS: Good point, I've removed it and refactored the tests to reflect.

lambdas/tests/unit/handlers/test_nhs_oauth_token_generator_handler.py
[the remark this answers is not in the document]
AndyFlintNHS: Both great suggestions. I've added that.

lambdas/tests/unit/services/base/test_nhs_oauth_service.py
AndyFlintNHS: You mentioning this actually made me realise that the nhs_oauth_response method wasn't needed. The only thing that used the get_new_access_token was get_active_access_token, which could just pick the token out of the response. I've refactored all of this now.

lambdas/services/base/nhs_oauth_service.py
```diff
@@ -24,10 +24,11 @@ def get_active_access_token(self):
             int(access_token_response["expires_in"])
             + int(access_token_response["issued_at"]) / 1000
         )
+
         time_safety_margin_seconds = 10
         remaining_time_before_expiration = access_token_expiration - time.time()
         if remaining_time_before_expiration < time_safety_margin_seconds:
-            access_token = self.get_new_access_token()
+            access_token = self.get_new_access_token().get("access_token", "")
```
NogaNHS, line 31: I think the function name change was right. As `get_new_access_token` is not returning only access_token

NogaNHS (approved):

abbas-khan10 (approved):

PR #585, opened 2025-03-24
[PRMP-1699] Replace isBSOL with ICB ODS code

NogaNHS:


lambdas/enums/repository_role.py
```diff
@@ -13,6 +13,7 @@ def list() -> List[str]:
         return [str(item.value) for item in RepositoryRole]
 
 
-class OrganisationRelationship(Enum):
-    BSOL_REL_ID = "RE4"
+class OrganisationRelationship(StrEnum):
+    COMMISSIONED_BY = "RE4"
     BSOL_ORG_ID = "15E"
```
chrisbloe-nhse, line 18: Do we want to rename this to `BSOL_ICB_ODS_CODE`?

lambdas/services/ods_api_service.py
```diff
@@ -59,30 +59,30 @@ def fetch_organisation_with_permitted_role(self, ods_code_list: list[str]) -> Di
 
         if pcse_ods is not None:
             logger.info(f"ODS code {ods_code} is a PCSE, returning org data")
```
chrisbloe-nhse, line 61: `is PCSE`?

lambdas/services/ods_api_service.py
```diff
@@ -107,18 +107,20 @@ def find_and_get_pcse_ods(ods_code):
...
+                == OrganisationRelationship.CLINICAL_COMMISSIONING_GROUP
             ):
-                return True
+                icb_ods_code = rel["Target"]["OrgId"]["extension"]
+                logger.info(f"Found ICB code for user: {icb_ods_code}")
+                return icb_ods_code
     except (KeyError, TypeError):
         logger.info("Failure fetching relationships")
-    return False
+    return "No ICB code found"
```
abbas-khan10, line 126: Would it be better to return a None or empty string here so we can falsy check if we need to, otherwise we would have to make a conditional check against a specific string? If we're not expecting this to occur we could also log and raise exception here?

lambdas/services/ods_api_service.py
```diff
@@ -59,30 +59,30 @@ def fetch_organisation_with_permitted_role(self, ods_code_list: list[str]) -> Di
...
+def parse_ods_response(org_data, role_code, icb_ods_code) -> dict:
     org_name = org_data["Organisation"]["Name"]
     org_ods_code = org_data["Organisation"]["OrgId"]["extension"]
 
     response_dictionary = {
         "name": org_name,
         "org_ods_code": org_ods_code,
         "role_code": role_code,
-        "is_BSOL": is_bsol,
+        "icb_ods_code": icb_ods_code,
```
abbas-khan10, line 85: If this is our lambda response we should probably make this JSON?

lambdas/services/ods_api_service.py
NogaNHS: We only use this for logs, no checks needed against this string.

lambdas/enums/repository_role.py
NogaNHS: Done

lambdas/services/ods_api_service.py
NogaNHS: Done

abbas-khan10 (approved):

jack-nhs (approved):

PR #593, opened 2025-03-26
[PRMP-1504] NRL pointer check before creation

NogaNHS:


lambdas/services/nrl_api_service.py
```diff
@@ -41,8 +41,19 @@ def get_end_user_ods_code(self):
             ssm_key_parameter, with_decryption=True
         )
 
-    def create_new_pointer(self, body: dict, retry_on_expired: bool = True):
+    def create_new_pointer(
+        self,
+        nhs_number,
+        body: dict,
```
abbas-khan10, line 47: If this body is always going to be passed a `document.model_dump()` should we maybe use the data type `DocumentReferenceInfo` instead of dict?

lambdas/services/nrl_api_service.py
```diff
@@ -41,8 +41,19 @@ def get_end_user_ods_code(self):
             ssm_key_parameter, with_decryption=True
         )
 
-    def create_new_pointer(self, body: dict, retry_on_expired: bool = True):
+    def create_new_pointer(
+        self,
+        nhs_number,
```
abbas-khan10, line 46: Can we add data types in the function calls for this file? There's a few missing across several functions including this one?

lambdas/tests/unit/services/test_nrl_api_service.py
```diff
@@ -11,34 +11,115 @@
...
     nrl_service = NrlApiService(fake_ssm_service, fake_auth_service)
     mocker.patch.object(nrl_service, "session")
     yield nrl_service
 
 
-def test_create_new_pointer(nrl_service):
-    mock_body = {"test": "tests"}
+def test_create_new_pointer(nrl_service, mocker):
+    nrl_service.get_pointer = mocker.MagicMock(return_value={})
+    nrl_service.session.post = mocker.MagicMock()
```
abbas-khan10, line 23: Can we create fixtures here instead of injecting mocker in our unit tests?

lambdas/services/nrl_api_service.py
NogaNHS: I'm not sure if it's always going to be document. we can create pointers for other types

lambdas/services/nrl_api_service.py
NogaNHS: Done

lambdas/tests/unit/services/test_nrl_api_service.py
NogaNHS: Done

lambdas/services/nrl_api_service.py
abbas-khan10: Maybe one for another time then, but we should minimise passing unstructured dicts around wherever we can

AndyFlintNHS (approved):

PR #629, opened 2025-05-09
[PRMT-72] Allow non-GP admittance to the NDR

MohammadIqbalAD-NHS:


lambdas/services/ods_api_service.py
```diff
@@ -51,32 +51,55 @@ def fetch_organisation_with_permitted_role(self, ods_code_list: list[str]) -> Di
...
 
 
 def parse_ods_response(org_data, role_code, icb_ods_code) -> dict:
-    org_name = org_data["Organisation"]["Name"]
-    org_ods_code = org_data["Organisation"]["OrgId"]["extension"]
+    org_name = ""
+    org_ods_code = ""
+
+    if org_data:
+        org_name = org_data["Organisation"]["Name"]
```
abbas-khan10, line 101: Can we warp this is a try/catch for KeyError? As this will raise exceptions if we try to access dict keys without them being present (aware this logic should have already been implemented)

Alternatively, we can do `org_name = org_data.get("Organisation", {}).get("Name")` where `org_name` will be `None` if the key does not exist and we can do a check on `None` instead using `if org_name: `

chrisbloe-nhse (commented):
ITOC use a range of ODS codes to test (not just one).

lambdas/utils/constants/ssm.py
```diff
@@ -3,4 +3,8 @@
 PCSE_USER_ROLE_CODE = "/auth/smartcard/role/pcse"
 
 GP_ORG_ROLE_CODE = "/auth/org/role_code/gpp"
+
 PCSE_ODS_CODE = "/auth/org/ods_code/pcse"
+ITOC_ODS_CODE = "/auth/org/ods_code/itoc"
+
```
abbas-khan10, line 9: Can we remove the whitespaces in this file?

lambdas/services/ods_api_service.py
MohammadIqbalAD-NHS: We have gone with the `.get` approach to default to empty strings if the indexes do not exist, as there are cases where that is expected.

lambdas/services/ods_api_service.py
```diff
@@ -51,11 +51,11 @@ def fetch_organisation_with_permitted_role(self, ods_code_list: list[str]) -> Di
         ods_code = ods_code_list[0]
         logger.info(f"ods_code selected: {ods_code}")
 
-        itoc_ods_code = token_handler_ssm_service.get_itoc_ods_code()
+        itoc_ods_codes = token_handler_ssm_service.get_itoc_ods_code()
```
chrisbloe-nhse, line 54: get_itoc_ods_codes()

MohammadIqbalAD-NHS: Done

lambdas/utils/constants/ssm.py
MohammadIqbalAD-NHS: Done

MohammadIqbalAD-NHS:
> ITOC use a range of ODS codes to test (not just one).

SSM parameter on Dev has now been modified from a String to a StringList, code modified accordingly.

lambdas/services/token_handler_ssm_service.py
```diff
@@ -93,40 +95,45 @@ def get_smartcard_role_pcse(self) -> list[str]:
...
-            [
-                GP_ORG_ROLE_CODE,
-            ]
-        )
-        response = [params.get(GP_ORG_ROLE_CODE)]
-        if None in response:
+    def get_gp_org_role_code(self) -> str:
+        logger.info("starting ssm request to retrieve GP organisation role code")
+        response = self.get_ssm_parameter(GP_ORG_ROLE_CODE)
+        if not response:
```
PedroSoaresNHS, line 101: can we swap the order here to :
if response return response
and then the rest of the logic:
logger.error(
                LambdaError.LoginGpOrgRoleCode.to_str(),
                {"Result": "Unsuccessful login"},
            )
            raise LoginException(500, LambdaError.LoginGpOrgRoleCode)

lambdas/services/token_handler_ssm_service.py
```diff
@@ -93,40 +95,45 @@ def get_smartcard_role_pcse(self) -> list[str]:
...
-            [
-                PCSE_ODS_CODE,
-            ]
-        )
-        response = [params.get(PCSE_ODS_CODE)]
-        if None in response:
+    def get_pcse_ods_code(self) -> str:
+        logger.info("starting ssm request to retrieve PCSE ODS code")
+        response = self.get_ssm_parameter(PCSE_ODS_CODE)
+        if not response:
```
PedroSoaresNHS, line 112: same here, changing the if logic

lambdas/services/ods_api_service.py
```diff
@@ -89,22 +110,25 @@ def parse_ods_response(org_data, role_code, icb_ods_code) -> dict:
...
-def find_and_get_gpp_org_code(org_details):
-    logger.info("Checking GPP Roles")
-    json_roles: List[Dict] = org_details["Organisation"]["Roles"]["Role"]
+def get_user_gp_org_role_code(org_data):
+    logger.info("Checking if GP organisation role is present")
+    json_roles: List[Dict] = org_data["Organisation"]["Roles"]["Role"]
 
-    org_role_codes = token_handler_ssm_service.get_org_role_codes()
+    gp_org_role_code = token_handler_ssm_service.get_gp_org_role_code()
     for json_role in json_roles:
```
PedroSoaresNHS, line 118: add org_data variable type in get_user_gp_org_role_code(
make method return boolean
if we return false instead of none then we can change
   if gp_org_role_code is not None:
to  if gp_org_role_code:

lambdas/services/ods_api_service.py
```diff
@@ -110,7 +110,7 @@ def parse_ods_response(org_data, role_code, icb_ods_code) -> dict:
     return response_dictionary
 
 
-def get_user_gp_org_role_code(org_data):
+def get_user_gp_org_role_code(org_data) -> Optional[str]:
```
PedroSoaresNHS, line 113: org_data needs a type

lambdas/services/ods_api_service.py
```diff
@@ -121,7 +121,7 @@ def get_user_gp_org_role_code(org_data):
     return None
 
 
-def get_user_primary_org_role_code(org_data):
+def get_user_primary_org_role_code(org_data) -> str:
```
PedroSoaresNHS, line 124: org_data type

lambdas/services/token_handler_ssm_service.py
```diff
@@ -98,35 +98,40 @@ def get_smartcard_role_pcse(self) -> list[str]:
     def get_gp_org_role_code(self) -> str:
         logger.info("starting ssm request to retrieve GP organisation role code")
         response = self.get_ssm_parameter(GP_ORG_ROLE_CODE)
-        if not response:
+
+        if response:
+            return response
+        else:
```
PedroSoaresNHS, line 104: else is not necessary, since the if has a return

lambdas/services/ods_api_service.py
```diff
@@ -89,25 +110,28 @@ def parse_ods_response(org_data, role_code, icb_ods_code) -> dict:
...
-        return ods_code
-    return None
+def get_user_primary_org_role_code(org_data: Dict[str, Any]) -> str:
+    logger.info("Checking if a primary organisation role is present")
+    json_roles: List[Dict] = org_data["Organisation"]["Roles"]["Role"]
+
+    for json_role in json_roles:
+        if "primaryRole" in json_role:
+            return json_role["id"]
+    return ""
```
abbas-khan10, line 131: We're returning None above and and empty string here. Either one would work for the existence check, but I would keep it consistent with the `Optional[str]`

PedroSoaresNHS (approved):
LGTM

lambdas/services/ods_api_service.py
MohammadIqbalAD-NHS: Since at later parts of the code the value returned from this function will be indexed and a string would be expected, this is why I opted for returning an empty string.

abbas-khan10 (approved):

PR #631, opened 2025-05-14
[PRMT-69] Metadata pre-processor lambda to handle variation 

abbas-khan10:


lambdas/handlers/bulk_upload_metadata_preprocessor_handler.py
```diff
@@ -0,0 +1,30 @@
...
+
+logger = LoggingService(__name__)
+
+
+@set_request_context_for_logging
+@override_error_check
+@ensure_environment_variables(names=["STAGING_STORE_BUCKET_NAME"])
+@handle_lambda_exceptions
+def lambda_handler(event, _context):
+    practice_directory = event.get("practiceDirectory")
```
MohammadIqbalAD-NHS, line 18: On production, would the value of "practiceDirectory" always be an ODS Code? If so, would it be better renaming this as 
```suggestion
    ods_code = event.get("odsCode")
```

lambdas/services/bulk_upload_metadata_preprocessor_service.py
```diff
@@ -0,0 +1,404 @@
...
+    def extract_document_path(
+        file_path: str,
+    ) -> tuple[str, str]:
+        document_number_expression = r"(.*[/])*((\d+)[^0-9]*of[^0-9]*(\d+)(.*))"
+
+        expression_result = regex.search(rf"{document_number_expression}", file_path)
+
+        if expression_result is None:
+            logger.info("Failed to find the document number in file name")
+            raise InvalidFileNameException("Incorrect document number format")
```
MohammadIqbalAD-NHS, line 320: This log/error needs modifying to indicate a failure to find the document path, rather than the document number.

lambdas/services/bulk_upload_metadata_preprocessor_service.py
```diff
@@ -0,0 +1,404 @@
...
+    def extract_lloyd_george_record_from_bulk_upload_file_name(
+        file_path: str,
+    ) -> tuple[str, str]:
+        _expression = r".*?ll[oO0οՕ〇]yd.*?ge[oO0οՕ〇]rge.*?rec[oO0οՕ〇]rd(.*)"
+        lloyd_george_record = regex.search(
+            rf"{_expression}", file_path, regex.IGNORECASE
+        )
+        if lloyd_george_record is None:
+            logger.info("Failed to find Lloyd George Record")
+            raise InvalidFileNameException("Invalid Lloyd_George_Record separator")
```
MohammadIqbalAD-NHS, line 285: This sounds like it has failed trying to find an actual record. Could we be more explicit, for example:
```suggestion
            logger.error("Failed to extract Lloyd George Record from file name")
            raise InvalidFileNameException("Invalid Lloyd_George_Record separator")
```

lambdas/services/bulk_upload_metadata_preprocessor_service.py
```diff
@@ -0,0 +1,404 @@
...
+
+        current_file_path = lloyd_george_record.group(1)
+
+        lloyd_george_record_text = "Lloyd_George_Record"
+
+        return lloyd_george_record_text, current_file_path
```
MohammadIqbalAD-NHS, line 291: We are hardcoding this string and returning it, and afterwards passing it into the function `assemble_valid_file_name()`. Assuming that all valid file names should have the string `"Lloyd_George_Record"`, surely we can hardcode it instead on _line 399_ and avoid passing the string around?

lambdas/services/bulk_upload_metadata_preprocessor_service.py
```diff
@@ -0,0 +1,404 @@
...
+        )
+
+        if expression_result is None:
+            logger.info("Failed to find the person name in the file name")
+            raise InvalidFileNameException("Invalid patient name")
+
+        name = expression_result.group(1)
+        current_file_path = expression_result.group(2)
+
+        return name, current_file_path
```
MohammadIqbalAD-NHS, line 273: `person` and `patient` are being used interchangeably, I would suggest the following refactor:
```suggestion
    def extract_patient_name_from_bulk_upload_file_name(
        file_path: str,
    ) -> tuple[str, str]:
        document_number_expression = r".*?([\p{L}][^\d]*[\p{L}])(.*)"
        expression_result = regex.search(
            rf"{document_number_expression}", file_path, regex.IGNORECASE
        )

        if expression_result is None:
            logger.info("Failed to find the patient name in the file name")
            raise InvalidFileNameException("Invalid patient name")

        patient_name = expression_result.group(1)
        current_file_path = expression_result.group(2)

        return patient_name, current_file_path
```

lambdas/services/bulk_upload_metadata_preprocessor_service.py
```diff
@@ -0,0 +1,404 @@
...
+                self.extract_lloyd_george_record_from_bulk_upload_file_name(
+                    current_file_name
+                )
+            )
+            patient_name, current_file_name = (
+                self.extract_person_name_from_bulk_upload_file_name(current_file_name)
+            )
+
+            if sum(c.isdigit() for c in current_file_name) != 18:
+                logger.info("Failed to find NHS or date")
```
MohammadIqbalAD-NHS, line 124:
```suggestion
                logger.info("Failed to find NHS number or date")
```

lambdas/services/bulk_upload_metadata_preprocessor_service.py
```diff
@@ -0,0 +1,404 @@
...
+                raise InvalidFileNameException("incorrect NHS number or date format")
```
MohammadIqbalAD-NHS, line 125: What value is this if statement actually providing, considering the fact that the subsequent functions that check for nhs number and date would raise error messages that are more explicit anyways?

lambdas/handlers/bulk_upload_metadata_preprocessor_handler.py
abbas-khan10: Sometimes are ODS codes, the latest one was marked as "OSS" for the scanning provider so this was named accordingly

lambdas/services/bulk_upload_metadata_preprocessor_service.py
PedroSoaresNHS: there could be cases where the user is providing 11 NHS numbers and 8 date digits
we want to make sure that there is no pollution from nhs number to date

lambdas/services/bulk_upload_metadata_preprocessor_service.py
PedroSoaresNHS: there could be cases where the user is providing 11 NHS numbers and 8 date digits we want to make sure that there is no pollution from nhs number to date

lambdas/services/bulk_upload_metadata_preprocessor_service.py
PedroSoaresNHS: yes we can, however I think separating the logic of getting/assembling arguments is preferable.
In this sense, if we want we keep a structure of whatever argument we need will come from the extract...
And in the end we just add them all up. This way we exclude the need to know logic of Lloyd_George_Record will be in every record looking exactly how it does from the method that concatenates values and instead keep that logic in the method that handles extracting the Lloyd_George_Record

MohammadIqbalAD-NHS (approved):

PR #635, opened 2025-05-19
[PRMT-72] Move PCSE repository role check to first

MohammadIqbalAD-NHS:


lambdas/services/login_service.py
```diff
@@ -165,6 +165,13 @@ def remove_used_state(self, state):
     def generate_repository_role(self, organisation: dict, smartcard_role: str):
         logger.info(f"Smartcard Role: {smartcard_role}")
 
+        if smartcard_role in self.token_handler_ssm_service.get_smartcard_role_pcse():
+            logger.info("PCSE: smartcard ODS identified")
```
abbas-khan10, line 169: Can we put this log message after the second check against the PCSE ods code before the return?

PedroSoaresNHS (approved):
LGTM

abbas-khan10 (approved):

PR #637, opened 2025-05-20
PRMT-71

PedroSoaresNHS:
Improve efficiency of LG download
The download process now utilise S3 streaming
each file should be streamed from S3 into a BytesIO object and written to a zip file

#PRMT-71
https://nhsd-jira.digital.nhs.uk/browse/PRMT-71

### DIFF
--- a/lambdas/services/base/s3_service.py
+++ b/lambdas/services/base/s3_service.py
@@ -160,3 +160,17 @@ class S3Service:
     def get_file_size(self, s3_bucket_name: str, object_key: str) -> int:
         response = self.client.head_object(Bucket=s3_bucket_name, Key=object_key)
         return response.get("ContentLength", 0)
+
+    def get_object_stream(self, bucket: str, key: str):
+        response = self.client.get_object(Bucket=bucket, Key=key)
+        return response["Body"]
+
+    def upload_file_obj(self, file_obj, s3_bucket_name: str, file_key: str):
+        try:
+            self.client.upload_fileobj(file_obj, s3_bucket_name, file_key)
+            logger.info(f"Uploaded file object to s3://{s3_bucket_name}/{file_key}")
+        except ClientError as e:
+            logger.error(
+                f"Failed to upload file object to s3://{s3_bucket_name}/{file_key} - {e}"
+            )
+            raise
--- a/lambdas/services/generate_document_manifest_zip_service.py
+++ b/lambdas/services/generate_document_manifest_zip_service.py
@@ -1,6 +1,6 @@
+import io
 import os
 import shutil
-import tempfile
 import zipfile
 
 from botocore.exceptions import ClientError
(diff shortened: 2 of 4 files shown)


lambdas/services/base/s3_service.py
```diff
@@ -160,3 +160,17 @@ def list_all_objects(self, bucket_name: str) -> list[dict]:
...
+
+    def upload_file_obj(self, file_obj, s3_bucket_name: str, file_key: str):
+        try:
+            self.client.upload_fileobj(file_obj, s3_bucket_name, file_key)
+            logger.info(f"Uploaded file object to s3://{s3_bucket_name}/{file_key}")
+        except ClientError as e:
+            logger.error(
+                f"Failed to upload file object to s3://{s3_bucket_name}/{file_key} - {e}"
+            )
+            raise
```
abbas-khan10, line 176: `raise e`

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -68,18 +87,19 @@ def get_file_bucket_and_key(self, file_location: str):
                 "Failed to parse bucket from file location string"
             )
 
-    def upload_zip_file(self):
-        logger.info("Uploading zip file to s3")
+    def upload_zip_file(self, zip_buffer: io.BytesIO):
+        zip_file_key = f"{self.zip_file_name}"
```
abbas-khan10, line 91: Why remove the log??

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -20,42 +20,61 @@ class DocumentManifestZipService:
...
+            return zip_buffer
+
+        except Exception as e:
             self.update_failed_status()
-            msg = f"{file_key} may reference missing file in s3 bucket: {file_bucket}"
-            logger.error(
-                f"{LambdaError.ZipServiceClientError.to_str()} {msg + str(e)}",
-                {"Result": "Failed to create document manifest"},
-            )
+            if isinstance(e, GenerateManifestZipException):
```
abbas-khan10, line 73: Why are we doing this extra check here for GenerateManifestZipException it's already raised on line 64 which is what we want, all exceptions from `utils.lambda_exceptions` will be caught and handled by the handler using the `@handle_lambda_exceptions` decorator and return an appropriate response to api gateway? 

Also what and what other exceptions do we expect here? A catch all `Exception` is too broad

lambdas/tests/unit/services/test_generate_document_manifest_zip_service.py
```diff
@@ -175,3 +156,70 @@ def test_update_processing_status(mock_service):
...
+            with zipf.open(file_name) as f:
+                assert f.read() == file_content
+
+    assert mock_copyfileobj.called
+
+
+def test_stream_zip_documents_raises_client_error(
+    mocker, mock_service, mock_s3_service
+):
+    from botocore.exceptions import ClientError
```
abbas-khan10, line 189: We have a mock client error in conftest.py

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -20,42 +20,61 @@ class DocumentManifestZipService:
...
+                    file_bucket, file_key = self.get_file_bucket_and_key(
+                        document_location
+                    )
+                    try:
+                        # Stream file from S3
+                        s3_object_stream = self.s3_service.get_object_stream(
+                            file_bucket, file_key
+                        )
+
+                        with s3_object_stream as stream:
```
abbas-khan10, line 54: Why are we using the `with as` statement here? The `get_object_stream` returns the body from the S3 api call, is the api call not closed when the response is returned?

app/src/helpers/requests/getPresignedUrlForZip.ts
```diff
@@ -7,7 +7,7 @@ import waitForSeconds from '../utils/waitForSeconds';
 import { DownloadManifestError } from '../../types/generic/errors';
 import { isRunningInCypress } from '../utils/isLocal';
 
-export const DELAY_BETWEEN_POLLING_IN_SECONDS = isRunningInCypress() ? 0 : 10;
+export const DELAY_BETWEEN_POLLING_IN_SECONDS = isRunningInCypress() ? 0 : 5;
```
abbas-khan10, line 10: Can we reduce this to 3 seconds? The pending -> processing should happen instantly so if there is a small delay, 3 seconds should catch it.

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -1,6 +1,6 @@
+import io
```
abbas-khan10, line 1: from io import BytesIO

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -20,84 +20,79 @@ class DocumentManifestZipService:
...
+
+        with zipfile.ZipFile(zip_buffer, "w", compression=zipfile.ZIP_DEFLATED) as zipf:
+            for document_location, document_name in documents.items():
+                file_bucket, file_key = self.get_file_bucket_and_key(document_location)
+                try:
+                    # Stream file from S3
+                    s3_object_stream = self.s3_service.get_object_stream(
+                        file_bucket, file_key
+                    )
+                    with zipf.open(document_name, mode="w") as zip_member:
```
abbas-khan10, line 48: After doing a bit of digging the streaming "Body" request returned from S3 will not automatically close itself, so will remain in our buffer till garbage collection. This may cause an issue as we increase the size of files we are zipping, so for safety we should close the connection once we read the body so that the s3 body is removed from memory. I believe this can look something like:

```
from contextlib import closing

with closing(s3_service.get_object_stream("lg-bucket-name", key)) as s3_object_stream, zf.open(document_name, "w") as member:
    shutil.copyfileobj(body, member, 64 * 1024)
```

app/src/helpers/requests/getPresignedUrlForZip.ts
```diff
@@ -33,9 +33,12 @@ const getPresignedUrlForZip = async (args: Args) => {
 
     const jobId = await requestJobId(args);
     let pendingCount = 0;
-
-    while (pendingCount < 3) {
-        await waitForSeconds(DELAY_BETWEEN_POLLING_IN_SECONDS);
+    let firstAttempt = true;
+    while (pendingCount < 10) {
+        if (!firstAttempt) {
```
adamwhitingnhs, line 38: Could remove the firstAttempt variable and change this to `if (pendingCount > 0)`

samFallowfieldNHS (approved):

MohammadIqbalAD-NHS:
Cypress E2E tests are failing.

NogaNHS (changes requested):
Missing unit tests for S3 service changes.

lambdas/services/base/s3_service.py
```diff
@@ -169,3 +169,17 @@ def list_all_objects(self, bucket_name: str) -> list[dict]:
     def get_file_size(self, s3_bucket_name: str, object_key: str) -> int:
         response = self.client.head_object(Bucket=s3_bucket_name, Key=object_key)
         return response.get("ContentLength", 0)
+
+    def get_object_stream(self, bucket: str, key: str):
+        response = self.client.get_object(Bucket=bucket, Key=key)
+        return response.get("Body")
```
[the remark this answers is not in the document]
PedroSoaresNHS: Removed duplicated logic

lambdas/services/generate_document_manifest_zip_service.py
```diff
@@ -20,84 +21,78 @@ class DocumentManifestZipService:
...
-            logger.error(
-                f"{LambdaError.ZipServiceClientError.to_str()} {msg + str(e)}",
-                {"Result": "Failed to create document manifest"},
-            )
-            raise GenerateManifestZipException(
-                status_code=500, error=LambdaError.ZipServiceClientError
-            )
+    def stream_zip_documents(self) -> BytesIO:
+        logger.info("Streaming and zipping documents in-memory")
+        documents = self.zip_trace_object.files_to_download  # Dict[str, str]
```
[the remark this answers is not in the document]
PedroSoaresNHS: removed

PR #645, opened 2025-05-29
[PRM-299] Update Node.js from v18.x to v24.x

AndyFlintNHS:
Jira ticket - https://nhsd-jira.digital.nhs.uk/browse/PRM-299

The version of "@types/node" is 22 as this is the latest version available! It is compatible with Node 24.

### DIFF
--- a/.github/workflows/base-vitest-test.yml
+++ b/.github/workflows/base-vitest-test.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


chrisbloe-nhse (changes requested):
`README.md` also needs an update on the Node version.

chrisbloe-nhse (approved):
Thanks for making those changes, Andy :)

PR #652, opened 2025-06-04
[NDR-43] Added Event ID for PDS Logs

robg-nhs:


lambdas/utils/audit_logging_setup.py
```diff
@@ -37,21 +38,21 @@ def error(self, message, custom_args: dict = None, *args, **kwargs):
...
     def critical(self, message, custom_args: dict = None, *args, **kwargs):
-        self.logger.critical(
-            message, extra={"custom_args": custom_args}, *args, **kwargs
-        )
+        self.logger.critical(message, extra={"custom_args": custom_args}, *args, **kwargs)
+
+    def info_with_correlation(self, message, tracking_id, custom_args: dict = None, *args, **kwargs):
+        correlation = f"{tracking_id}-{uuid.uuid4()}"
+        self.logger.info(f"[{correlation}] - {message}", extra={"custom_args": custom_args}, *args, **kwargs)
+        return correlation
```
NogaNHS, line 55: Call `info_in_correlation` here instead of repeating code

NogaNHS: Or potentially create one function with correlation as optional param. 
We are logging AWS correlation_id, is there a reason to create a custom one?

NogaNHS (approved):

adamwhitingnhs (approved):

PR #668, opened 2025-06-23
[PRMT-229] Renamed updated bulk upload workflow to performance bulk upload

PedroSoaresNHS:
Renamed updated bulk upload to performance bulk-upload.yml
renamed PDS_FHIR_ALWAYS_TRUE to BYPASS_PDS


lambdas/utils/utilities.py
```diff
@@ -72,7 +72,7 @@ def get_pds_service() -> PatientSearch:
         return PdsApiService(ssm_service, auth_service)
     else:
         return MockPdsApiService(
-            always_pass_mock=os.getenv("PDS_FHIR_ALWAYS_TRUE") in ["True", "true"]
+            always_pass_mock=os.getenv("BYPASS_PDS") in ["True", "true"]
```
abbas-khan10, line 75: `always_pass_mock` isn't clear, what is the purpose of this param?

abbas-khan10: Also can set to os.getenv("BYPASS_PDS", "false").lower() == "true"

MohammadIqbalAD-NHS (approved):

PR #671, opened 2025-06-30
[PRMT-435] Stitching manual trigger doesn't add all records to queue

PedroSoaresNHS:
Updated the way files were being added to the Queue to use batches of 10 over individual entries
updated the way records where being extracted from dynamo since we were reaching dynamo limits


lambdas/services/base/sqs_service.py
```diff
@@ -16,6 +18,35 @@ def send_message_fifo(self, queue_url: str, message_body: str, group_id: str):
     def send_message_standard(self, queue_url: str, message_body: str):
         self.client.send_message(QueueUrl=queue_url, MessageBody=message_body)
 
+    def send_message_batch_standard(self, queue_url: str, messages: list[str]):
```
abbas-khan10, line 21: Can we make the delay configurable as a parameter? We don't want to add delay to everything that might use batched messages from the SQS service, only when needed

lambdas/services/base/sqs_service.py
```diff
@@ -16,6 +18,35 @@ def send_message_fifo(self, queue_url: str, message_body: str, group_id: str):
...
+                    "DelaySeconds": delay,
+                }
+            )
+
+        response = self.client.send_message_batch(
+            QueueUrl=queue_url,
+            Entries=entries,
+        )
+
+        if "Failed" in response and response["Failed"]:
```
abbas-khan10, line 39: Better practice to use `response.get("Failed")` instead for retrieving dict values, unless you can guarantee they exist otherwise you risk raising a `KeyError`

lambdas/services/mock_pds_service.py
```diff
@@ -36,10 +36,18 @@ def pds_request(self, nhs_number: str, *args, **kwargs) -> Response:
...
         response = Response()
         if self.always_pass_mock:
-            pds_patient_index = 3
-            pds_patient = mock_pds_results[pds_patient_index]
+            mock_file_name = "pds_patient_9000000068_M85143_gp.json"
+            file_path = os.path.join(
+                parent_dir_of_this_file, "services", "mock_data", mock_file_name
+            )
+            try:
+                with open(file_path) as f:
```
abbas-khan10, line 44: Can we use clearer var names here instead of `f`

lambdas/services/pdf_stitching_service.py
```diff
@@ -28,6 +30,15 @@
 logger = LoggingService(__name__)
 
 
+def batch(iterable, size):
```
abbas-khan10, line 33: missing types in params

lambdas/services/base/sqs_service.py
PedroSoaresNHS: done

lambdas/services/pdf_stitching_service.py
PedroSoaresNHS: done

lambdas/services/pdf_stitching_service.py
```diff
@@ -22,12 +25,22 @@
...
 
 logger = LoggingService(__name__)
 
 
+def batch(iterable: Iterable[str], size: int) -> Iterator[List[str]]:
+    it = iter(iterable)
+    while True:
+        chunk = list(islice(it, size))
+        if not chunk:
+            break
```
abbas-khan10, line 40: This function is out of place, if it belongs to the PdfStitchingService, then put it into the class, if it's reusable helper then it can go into utils

lambdas/enums/lambda_error.py
```diff
@@ -8,26 +8,22 @@
...
         message = self.value["message"]
         if "%" in message and params:
             message = message % params
         interaction_id = getattr(request_context, "request_id", None)
         error_response = ErrorResponse(
-            err_code=err_code,
-            message=message,
-            interaction_id=interaction_id,
-            **kwargs
+            err_code=err_code, message=message, interaction_id=interaction_id, **kwargs
```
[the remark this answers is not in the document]
PedroSoaresNHS: accepted

lambdas/services/base/sqs_service.py
```diff
@@ -16,6 +18,26 @@ def send_message_fifo(self, queue_url: str, message_body: str, group_id: str):
...
+        entries = []
+        for i, body in enumerate(messages):
+            delay = base_delay
+            entries.append(
+                {
+                    "Id": str(uuid.uuid4()),
+                    "MessageBody": body,
+                    "DelaySeconds": delay,
+                }
+            )
```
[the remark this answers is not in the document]
PedroSoaresNHS: done

MohammadIqbalAD-NHS (approved):

abbas-khan10 (approved):

PR #673, opened 2025-07-02
[PRM-300] NDR - remove build warnings

jack-nhs:


chrisbloe (approved):
Query about use of `@types/sinon` otherwise the output and tests are as expected.

steph-torres-nhs (approved):

PR #675, opened 2025-07-03
[PRMT-70] Improve efficiency of LG retrieval

PedroSoaresNHS:
Fixed type issue

### DIFF
--- a/lambdas/services/pdf_stitching_service.py
+++ b/lambdas/services/pdf_stitching_service.py
@@ -3,7 +3,6 @@ import re
 import sys
 import uuid
 from datetime import datetime, timezone
-from decimal import Decimal
 from io import BytesIO
 
 from botocore.exceptions import ClientError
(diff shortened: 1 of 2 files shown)


lambdas/services/pdf_stitching_service.py
```diff
@@ -119,7 +120,7 @@ def create_stitched_reference(
                 "file_location": f"s3://{self.target_bucket}/{document_reference.nhs_number}/{reference_id}",
                 "file_name": f"1of1_{stripped_filename}",
                 "file_size": stitch_file_size,
-                "last_updated": date_now.timestamp(),
+                "last_updated": Decimal(str(date_now.timestamp())),
```
abbas-khan10, line 123: Is there a unit test potentially missing here as we didn't catch the issue in main?

PedroSoaresNHS: python date time comes as float, and the change in main made it so that it needs to be passed as decimal.
We can add a Unit test for it, but it probably is overkill

abbas-khan10: This field is an `int` in the model, so why cast to a Decimal?

PedroSoaresNHS: made into int

abbas-khan10 (approved):

samFallowfieldNHS (approved):

PR #721, opened 2025-08-05
[PRMT-523] Analysis on Lloyd George use case migration to defined Document Metadata

PedroSoaresNHS:
Analysis on Lloyd George use case migration to defined Document Metadata

https://nhsd-jira.digital.nhs.uk/browse/PRMT-523

### DIFF
--- a/lambdas/scripts/dynamodb_migration_20250731.py
+++ b/lambdas/scripts/dynamodb_migration_20250731.py
@@ -39,9 +39,6 @@ class VersionMigration:
                 all_entries=all_entries, total_count=total_count
             )
             self.run_status_migration(all_entries=all_entries, total_count=total_count)
-            self.run_uploading_migration(
-                all_entries=all_entries, total_count=total_count
-            )
         except Exception as e:
             self.logger.error("Migration failed", exc_info=e)
             raise
@@ -144,14 +141,6 @@ class VersionMigration:
             all_entries,
             total_count,
             lambda e: {"Status": "current"} if e.get("Status") != "current" else None,
-        )
-
-    def run_uploading_migration(self, all_entries, total_count):
-        self.process_entries(
-            "Uploading",
-            all_entries,
-            total_count,
-            lambda e: {"Uploading": False} if e.get("Uploading") is not False else None,
         )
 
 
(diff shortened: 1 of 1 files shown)


lambdas/scripts/dynamodb_migration_20250731.py
```diff
@@ -0,0 +1,176 @@
...
+import logging
+import sys
+
+from services.base.dynamo_service import DynamoDBService
+
+
+class VersionMigration:
+    def __init__(self, environment: str, table_name: str, dry_run: bool = False):
+        self.environment = environment
+        self.table_name = table_name
```
NogaNHS, line 11: This is only being used in line 16

lambdas/scripts/dynamodb_migration_20250731.py
```diff
@@ -0,0 +1,176 @@
...
+    def run_author_migration(self, all_entries, total_count):
+        self.logger.info("Running author migration")
+
+        bulk_reports = self.dynamo_service.scan_whole_table(self.bulk_upload_table)
+        report_lookup = {
+            (r.get("NhsNumber"), r.get("PdsOdsCode")): r.get("UploaderOdsCode")
+            for r in bulk_reports
+        }
+
+        def author_update(entry):
```
danielleandersonNHS, line 95: The author should be populated from the most recent complete upload in the report table? Doesn't take into account failures and replacements

lambdas/scripts/dynamodb_migration_20250731.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def run_deleted_cleanup_migration(self, all_entries, total_count):
+        self.process_entries(
+            "Deleted",
+            all_entries,
+            total_count,
+            lambda e: {"Deleted": ""} if e.get("Deleted") == "<empty>" else None,
+        )
+
+    def run_custodian_migration(self, all_entries, total_count):
```
danielleandersonNHS, line 112: Need to consider if the CurrentGpOds will long term be maintained. It doesn't align to the metadata but may still be maintained for reporting and some of the API work. If required this should be explicitly called out as maintain outside of the standard metadata fields

PR #761, opened 2025-08-20
[PRMT-574] Create a new version of Bulk Upload

ghost:
https://nhsd-jira.digital.nhs.uk/browse/PRMT-574

### DIFF
--- a/.github/workflows/base-lambdas-reusable-deploy-all.yml
+++ b/.github/workflows/base-lambdas-reusable-deploy-all.yml
@@ -148,6 +148,7 @@ jobs:
 
   deploy_v2_bulk_upload_metadata_lambda:
     name: Deploy v2_bulk_upload_metadata_lambda
+    if: ${{ (inputs.environment != 'prod') && (inputs.environment != 'pre-prod') }}
     uses: ./.github/workflows/base-lambdas-reusable-deploy.yml
     with:
       environment: ${{ inputs.environment}}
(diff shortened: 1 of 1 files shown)


.github/workflows/base-lambdas-reusable-deploy-all.yml
```diff
@@ -146,6 +146,20 @@ jobs:
...
+      environment: ${{ inputs.environment}}
+      python_version: ${{ inputs.python_version }}
+      build_branch: ${{ inputs.build_branch}}
+      sandbox: ${{ inputs.sandbox }}
+      lambda_handler_name: bulk_upload_metadata_handler
+      lambda_aws_name: V2BulkUploadMetadataLambda
+      lambda_layer_names: 'core_lambda_layer'
+    secrets:
+      AWS_ASSUME_ROLE: ${{ secrets.AWS_ASSUME_ROLE }}
+
```
NogaNHS, line 162: Is this going to pre-prod and prod?

.github/workflows/performance-base-run-bulk-upload.yml
```diff
@@ -38,6 +38,11 @@ on:
         required: false
         type: "string"
         default: "1"
+      use_v2_process:
+        description: "Use V2 Bulk Upload Process"
+        required: true
+        type: "boolean"
+        default: false
```
NogaNHS, line 45: Why not just deploy V2 to the original lambda if the flag is on? why do you need 2 lambdas?

.github/workflows/base-lambdas-reusable-deploy-all.yml
MohammadIqbalAD-NHS: For the time being up to Dev, will not go to Prod.

.github/workflows/performance-base-run-bulk-upload.yml
MohammadIqbalAD-NHS: As there is a need to have a separate V2 lambda for testing on Dev, we have avoided anything that uses/modifies existing resource behaviours.

PR #767, opened 2025-08-27
[PRME-189] Add handling for backwards navigation on upload pages

adamwhitingnhs:


app/src/components/blocks/_documentUpload/documentUploadCompleteStage/DocumentUploadCompleteStage.tsx
```diff
@@ -6,17 +6,29 @@ import usePatient from '../../../../helpers/hooks/usePatient';
...
     const nhsNumber: string = patientDetails?.nhsNumber ?? '';
     const formattedNhsNumber = formatNhsNumber(nhsNumber);
-    const dob: string = getFormattedDateFromString(patientDetails?.birthDate)
+    const dob: string = getFormattedDateFromString(patientDetails?.birthDate);
     const patientName = getFormattedPatientFullName(patientDetails);
 
     useTitle({ pageTitle: 'Record upload complete' });
 
+    if (documents.some((doc) => doc.state !== DOCUMENT_UPLOAD_STATE.SUCCEEDED)) {
+        navigate(routes.HOME);
```
lillie-dae, line 29: Presume this check only happens if documents fail for some reason?
Do we need a page "something went wrong" or something for this?

lillie-dae (approved):

kamenbachvarov-nhs (approved):

app/src/components/blocks/_documentUpload/documentUploadCompleteStage/DocumentUploadCompleteStage.tsx
adamwhitingnhs: We should only be showing the complete stage if all files are set to succeeded, so this handles if the user clicks back after they navigate away from here.

PR #779, opened 2025-09-17
[PRMT-617] Implementing consistent pagination when querying DynamoDB

NogaNHS:


MohammadIqbalAD-NHS (dismissed):

lambdas/services/manage_user_session_access.py
```diff
@@ -22,14 +22,14 @@ def find_login_session(self, ndr_session_id: str):
...
-        query_response = self.db_service.query_all_fields(
+        query_response = self.db_service.query_table(
             table_name=self.session_table_name,
             search_key="NDRSessionId",
             search_condition=ndr_session_id,
         )
 
         try:
-            current_session = query_response["Items"][0]
+            current_session = query_response[0]
```
SanjayChopraNHS, line 32: is it worth checking for if not query_response - raise exception so its more explicit and then set current_session

lambdas/services/base/dynamo_service.py
```diff
@@ -37,15 +36,14 @@ def get_table(self, table_name):
...
         self,
         table_name,
-        index_name,
         search_key,
         search_condition: str,
+        index_name: str = None,
         requested_fields: list[str] = None,
         query_filter: Attr | ConditionBase = None,
-        exclusive_start_key: dict = None,
     ):
```
SanjayChopraNHS, line 47: Could maybe have return type hint -> list[dict]

SanjayChopraNHS (approved):
minor thoughts/comments

MohammadIqbalAD-NHS (approved):

PR #784, opened 2025-09-19
[PRME-296] Add missing function return types

adamwhitingnhs:
This PR adds function return types every where they are missing, as well as replacing all non-arrow functions with arrow functions.


kamenbachvarov-nhs:
I see some return types are still missing in the _arf block. Are they out of scope ?

kamenbachvarov-nhs (approved):

PR #786, opened 2025-09-24
[NDR-244] Add E2E test readme instructions

megan-bower4:
Add setup instructions to README for E2E tests.


robg-test (changes requested):
Can we move this to within the folder containing the E2e Tests?

PR #787, opened 2025-09-24
[PRMP-189] Update new metadata lambda to add filename corrections to the queue

PedroSoaresNHS:
Updated message queue format to accept a new message format 
Updated metadata processor service to send the new message format to the queue

https://nhsd-jira.digital.nhs.uk/browse/PRMP-189


lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
```diff
@@ -206,9 +226,27 @@ def test_download_metadata_from_s3_raise_error_when_failed_to_download(
...
+        "/1234567890/2of2_Lloyd_George_Record_[Joe Bloggs]_[1234567890]_[25-12-2019].pdf,"
+        "/1234567890/2of2_Lloyd_George_Record_[Joe Bloggs]_[1234567890]_[25-12-2019].pdf,,"
+        "Y12345,LG,,03/09/2022,NEC,NEC,04/10/2023,1234567890\n"
+        "1of1_Lloyd_George_Record_[Joe Bloggs_invalid]_[123456789]_[25-12-2019].txt,"
+        "1of1_Lloyd_George_Record_[Joe Bloggs_invalid]_[123456789]_[25-12-2019].txt,,"
+        "Y12345,LG,,04/09/2022,NEC,NEC,04/10/2023,123456789\n"
+        "1of1_Lloyd_George_Record_[Jane Smith]_[1234567892]_[25-12-2019].txt,"
+        "1of1_Lloyd_George_Record_[Jane Smith]_[1234567892]_[25-12-2019].txt,,"
+        "Y12345,LG,,04/09/2022,NEC,NEC,04/10/2023,\n"
+    )
```
MohammadIqbalAD-NHS, line 244: Would it be better to have this test data in a dedicated CSV file, and then refer to its content in the test?

lambdas/models/staging_metadata.py
```diff
@@ -19,6 +19,7 @@ class MetadataFile(BaseModel):
     )
 
     file_path: str = Field(alias="FILEPATH")
+    stored_file_name: Optional[str] = Field(alias="STORED-FILE-NAME", default=None)
```
MohammadIqbalAD-NHS, line 22: As stored_file_name should always have a value, it should not be Optional, nor have a default.
```suggestion
    stored_file_name: str = Field(alias="STORED-FILE-NAME")
```

lambdas/models/staging_metadata.py
```diff
@@ -1,6 +1,6 @@
 from typing import Optional
 
-from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, field_validator
+from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, field_validator, model_validator
```
NogaNHS, line 3: Do you use `model_validator`?

lambdas/models/staging_metadata.py
```diff
@@ -19,6 +19,7 @@ class MetadataFile(BaseModel):
     )
 
     file_path: str = Field(alias="FILEPATH")
+    stored_file_name: str = Field(alias="STORED-FILE-NAME")
```
NogaNHS, line 22: This model is also used to when reading from the metadata csv, before bulk upload. I think this would break the metadata lambda. Have you testes the full bulk upload process with this change?

lambdas/tests/unit/services/test_bulk_upload_metadata_service.py
PedroSoaresNHS: we have a test that tests opening a file (test_csv_to_staging_metadata_raise_error_when_metadata_invalid)
this test tests processing the information, so keeping it like this it stays as a unit test instead of as an integration test

PR #789, opened 2025-09-26
[PRMP-198] Created author migration

PedroSoaresNHS:
Migration to meta data standards, by updating author column

https://nhsd-jira.digital.nhs.uk/browse/PRMP-198


PedroSoaresNHS:
the code will need to change, but it is like this now so that it is testable

lambdas/scripts/dynamodb_migration_prmp_198.py
```diff
@@ -0,0 +1,160 @@
...
+    def main(self):
+        self.logger.info("Starting version migration")
+        self.logger.info(f"Target table: {self.target_table}")
+        self.logger.info(f"Dry run mode: {self.dry_run}")
+
+        try:
+            all_entries = self.dynamo_service.scan_whole_table(
+                table_name=self.target_table
+            )
+            total_count = len(all_entries)
```
robg-test, line 28: Will this handle a large dataset?

Be useful to run this on `ndr-test`

lambdas/scripts/dynamodb_migration_prmp_198.py
```diff
@@ -0,0 +1,160 @@
...
+                )
+            else:
+                self.logger.info(f"Updating item {item_id} with {updated_fields}")
+                self.dynamo_service.update_item(
+                    table_name=self.target_table,
+                    key_pair={"ID": item_id},
+                    updated_fields=updated_fields,
+                )
+
+        self.logger.info(f"{label} migration completed.")
```
robg-test, line 76: Can we add within the `else` statement since you'll hit this at the moment on a dry run

lambdas/scripts/dynamodb_migration_prmp_198.py
```diff
@@ -0,0 +1,160 @@
...
+
+        bulk_upload_report_lookup: dict[str, dict] = {}
+
+        for row in bulk_reports:
+            nhs = row.get("NhsNumber")
+            status = row.get("UploadStatus")
+            timestamp = row.get("Timestamp")
+
+            if not nhs or status != "complete" or not timestamp:
+                continue
```
robg-test, line 99: Do we need an exception here? If something is missing any of these I think we might need to know

lambdas/scripts/dynamodb_migration_prmp_198.py
```diff
@@ -0,0 +1,160 @@
...
+
+            stored_report = bulk_upload_report_lookup.get(nhs)
+            if not stored_report or int(timestamp) > int(
+                stored_report.get("Timestamp", 0)
+            ):
+                bulk_upload_report_lookup[nhs] = row
```
robg-test, line 105:
```suggestion
            if not stored_report:
                   bulk_upload_report_lookup[nhs] = row
            elif int(timestamp) > int(stored_report.get("Timestamp", 0)):
                   bulk_upload_report_lookup[nhs] = row
```

I've made this suggestion for readability purposes, The first if is if you don't have a report, the second if being if the entry is newer. It's a bit against DRY but might be easier to understand?

lambdas/scripts/dynamodb_migration_prmp_198.py
```diff
@@ -0,0 +1,160 @@
...
+            if not bulk_upload_row:
+                self.logger.warning(
+                    f"No matching completed bulk upload found for NHS {nhs_number}"
+                )
+                return None
+
+            new_author = bulk_upload_row.get("UploaderOdsCode")
+            if not new_author:
+                self.logger.warning(f"No uploader ODS code found for NHS {nhs_number}")
+                return None
```
robg-test, line 124: I think we might need to have a think about what ends up on the DLQ and what we skip.
If we don't have the data and never feasibly get it back we have to skip?

lambdas/scripts/dynamodb_migration_prmp_198.py
SWhyteAnswer: @robg-test when this is actually use it will be scanning a segment I think this is just here untill the infrastructure is complete

lambdas/scripts/dynamodb_migration_prmp_198.py
PedroSoaresNHS: done

samfallowfield (dismissed):

SWhyteAnswer (dismissed):

PR #793, opened 2025-09-29
[PRMP-202] Migration to Data Standards - LG table values

PedroSoaresNHS:
Added migration for:
Custodian,
Status,
Snomed code
Doc status

Also changed how the entries are retrieved from the table, so that they do not stay all in memory.
https://nhsd-jira.digital.nhs.uk/browse/PRMP-202


lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,158 @@
...
+                )
+                self.logger.info(f"{label} migration completed.")
+
+
+    @staticmethod
+    def update_custodian_entry(entry: dict) -> dict | None:
+        custodian = entry.get("Custodian")
+        current_gp_ods = entry.get("CurrentGpOds")
+
+        if not custodian and current_gp_ods:
```
robg-test, line 85: Think we might need to change this.

If the `custodian` is missing, take `current_gp_ods`
If `custodian` is present but differs from `current_gp_ods` take `current_gp_ods`
If `custodian` is the same as `current_gp_ods` we can skip

Check with @bethany-kish-nhs on this one, I think this comes back to the migration of the DECE, SUSP and REST data in CurrentGpODS

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,156 @@
...
+        description="Migrate DynamoDB table columns",
+    )
+    parser.add_argument("environment", help="Environment prefix for DynamoDB table")
+    parser.add_argument("table_name", help="DynamoDB table name to migrate")
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Run migration in dry-run mode (no writes)",
+    )
+    args = parser.parse_args()
```
NogaNHS, line 141: As this is a lambda, do we need all of this?

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,156 @@
...
+
+def setup_logging():
+    importlib.reload(logging)
+    logging.basicConfig(
+        level=logging.INFO,
+        format="[%(asctime)s] %(levelname)s [%(name)s.%(funcName)s:%(lineno)d] %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+        stream=sys.stdout,
+    )
+
```
NogaNHS, line 125: Any reason not to use our logging class?

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,156 @@
...
+    @staticmethod
+    def update_status_entry(entry: dict) -> dict | None:
+        if entry.get("Status") != "current":
+            return {"Status": "current"}
+        return None
+
+    @staticmethod
+    def update_document_snomed_code_type_entry(entry: dict) -> dict | None:
+        fixed_snomed_code = 16521000000101
```
NogaNHS, line 95: Please use the Enum here

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,156 @@
...
+        if entry.get("DocStatus"):
+            return None
+
+        item_id = entry.get("ID")
+
+        if entry.get("Deleted"):
+            return {"DocStatus": "deprecated"}
+        if entry.get("Uploaded"):
+            return {"DocStatus": "final"}
+        if entry.get("Uploading"):
```
NogaNHS, line 110: Would be good to use the `DocumentReference` model

lambdas/scripts/dynamodb_migration_prmp_202.py
PedroSoaresNHS: We do not,
But it is the only way to test that the migration works.
So once the lambda and step functions are created, this will be removed

lambdas/scripts/dynamodb_migration_prmp_202.py
PedroSoaresNHS: removing this logger

lambdas/scripts/dynamodb_migration_prmp_202.py
PedroSoaresNHS: done

lambdas/scripts/dynamodb_migration_prmp_202.py
SWhyteAnswer: @bethany-kish-nhs can you have a look at this?

lambdas/models/document_reference.py
```diff
@@ -170,3 +170,12 @@ def set_virus_scanner_result(self, updated_virus_scanner_result) -> None:
...
     def set_uploaded_to_true(self):
         self.uploaded = True
+
+    def infer_doc_status(self) -> str | None:
+        if self.deleted:
+            return "deprecated"
+        if self.uploaded:
+            return "final"
+        if self.uploading:
+            return "preliminary"
```
NogaNHS, line 180: What if both `uploaded` and `uploading` are true?

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,192 @@
...
+        self.logger.info(f"Target table: {self.target_table}")
+        self.logger.info(f"Dry run mode: {self.dry_run}")
+
+        if entries is None:
+            self.logger.info("No entries provided — scanning entire table.")
+            raise ValueError("Entries must be provided to main().")
+
+        return [
+            ("LGTableValues", self.update_entry)
+        ]
```
NogaNHS, line 38: Do we need this?

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,192 @@
...
+                    updated_fields=updated_fields,
+                )
+        self.logger.info(f"{label} migration completed.")  # Moved outside the loop
+
+    def update_entry(self, entry: dict ) -> dict | None:
+        """
+        Aggregates updates from all update methods for a single entry.
+        Returns a dict of fields to update, or None if no update is needed.
+        """
+        updates = {}
```
NogaNHS, line 83: maybe rename this too

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,192 @@
...
+
+        custodian_update = self.update_custodian_entry(entry)
+        if custodian_update:
+            updates.update(custodian_update)
+
+        status_update = self.update_status_entry(entry)
+        if status_update:
```
NogaNHS, line 90:
```suggestion
        if status_update := self.update_status_entry(entry)
```

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,191 @@
...
+        Main entry point for the migration.
+        Returns a list of (label, update function) tuples.
+        Accepts a list of entries for Lambda-based execution, or scans the table if `entries` is None.
+        """
+        self.logger.info("Starting version migration")
+        self.logger.info(f"Target table: {self.target_table}")
+        self.logger.info(f"Dry run mode: {self.dry_run}")
+
+        if entries is None:
+            self.logger.info("No entries provided — scanning entire table.")
```
MohammadIqbalAD-NHS:
```suggestion
            self.logger.error("No entries provided after scanning entire table.")
```

lambdas/scripts/dynamodb_migration_prmp_202.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, environment: str, table_name: str, dry_run: bool = False):
+        self.environment = environment
+        self.table_name = table_name
+        self.dry_run = dry_run
+        self.logger = logging.getLogger("VersionMigration")
+        self.dynamo_service = DynamoDBService()
+
+        self.target_table = f"{self.environment}_{self.table_name}"
+
+    def main(
```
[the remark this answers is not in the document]
PedroSoaresNHS: no reason in particular, we just need a generic name since we want every migration to have the same initial contact point.
This way our lambda can be generic

MohammadIqbalAD-NHS (approved):

NogaNHS (approved):

PR #798, opened 2025-10-01
[NDR-251] Allow PDM routing in a POST DocumentReference

jameslinnell:
Allows POSTing to PDM routing.
- Determine client from the "common name" in the MTLS header.
- Set the SNOMED code to PATIENT_DATA if coming through mTLS and common name == pdm
- Set the database to PDM based on the SNOMED code
- Set the database based on mapping rather than a conditional. 
- Unit Tests

### DIFF
--- a/lambdas/enums/mtls.py
+++ b/lambdas/enums/mtls.py
@@ -1,0 +1,5 @@
+from enum import StrEnum, auto
+
+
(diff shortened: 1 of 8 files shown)


jameslinnell:
Moved to a different ticket

PR #799, opened 2025-10-01
[PRMP-179] Adding update function and get_s3_values

SWhyteAnswer:


lambdas/scripts/dynamodb_migration_prmp_179.py
```diff
@@ -0,0 +1,176 @@
...
+        if not s3_path or not s3_path.startswith("s3://"):
+            return None
+
+        path = s3_path[5:]
+        parts = path.split("/", 1)
+
+        if len(parts) != 2 or not parts[0] or not parts[1]:
+            return None
+
+        return parts[0], parts[1]
```
PedroSoaresNHS, line 98: probably better to assign them in the method, and return s3_bucket, s3_key
instead of returning s3_bucket_path_parts and then doing 
s3_bucket, s3_key = s3_bucket_path_parts

MohammadIqbalAD-NHS (dismissed):

lambdas/scripts/dynamodb_migration_prmp_179.py
```diff
@@ -0,0 +1,176 @@
+import argparse
+from typing import Iterable, Callable
+from services.base.dynamo_service import DynamoDBService
+from utils.audit_logging_setup import LoggingService
+from services.base.s3_service import S3Service
+
+class VersionMigration:
+    filesize_key = "FileSize"
+    s3_key = "S3Key"
```
NogaNHS: Can we change this to S3FileKey to match existing field in model?

PedroSoaresNHS (approved):

MohammadIqbalAD-NHS (approved):

PR #801, opened 2025-10-02
[PRMP-337] Refactor custodian assignment logic in document reference

NogaNHS:


MohammadIqbalAD-NHS:
> The PCSE Code in `ods_utils.py` is wrong. Check with @bethany-kish-nhs for the correct one

`X4S4L`, this is the ods code we have used consistently, I cannot see how it is incorrect.

